PR #44, opened 2024-09-10
Fix UCSC hyperlink markdown in about.mdx

nakucher:
Was missing closing parens for the hyperlink to the UCSC genome browser.


NoopDog (approved):
Thanks!

PR #48, opened 2024-09-11
feat: update genome page URL to remove the "." in the genome version id (#29)

### LINKED ISSUE #29: Update genome page URL to remove the "." in the genome version id
Make the URLs like
https://brc-analytics.dev.clevercanary.com/data/genomes/GCA_001189475.1

Looks like
https://brc-analytics.dev.clevercanary.com/data/genomes/GCA_001189475_1 

The "." is causing a problem with the next path mapper in cloudfront and we get a 404 if we try to go to the page directly.

frano-m:
Closes #29.


NoopDog (approved):
Thanks @frano-m 🚀

Issue #177, opened 2024-11-15
UI Exploration on Genomes list


NoopDog:
On datasets:

1. Change the link and label back to "Genomes" (for now)
2. Update content to come from NCBI as per #157
3. Update the columns and filters as per #157 
4. Add tags 
5. Keep the existing links to NCBI Taxa, NCBI genome page, UCSC Browser
6. Wait for workflow ids

PR #178, opened 2024-11-17
feat: rename genome list and update to use data from ncbi (#177)

hunterckx:
- ~~Tags are not included because they're associated with organisms, not genomes~~ -- Organisms are now included
- ~~The UCSC Browser URLs of uncertain accuracy are still included~~ -- I believe we've confirmed that our current matching method is accurate


files/build-files-from-ncbi.py
```diff
@@ -52,6 +53,38 @@ def get_genome_row(genome_info):
...
+    elif "augustus" in match:
+      augustus_file = match
+  if augustus_file:
+    return urllib.parse.urljoin(f"{url}/", augustus_file)
+  # No match, I guess that's OK ?
+  return None
+
+
+def add_gene_model_url(genomes_df: pd.DataFrame):
+  return pd.concat([genomes_df, genomes_df["accession"].apply(_id_to_gene_model_url).rename("geneModelUrl")], axis="columns")
```
hunterckx, line 86: I've added gene model URL -- not certain whether `accession` is appropriate for getting that

NoopDog:
@hunterckx can you rebase this on main please?

NoopDog:
Obsolete! Closing

PR #199, opened 2024-12-10
chore: update dev dependancies

jpaten:
### Changes
- Updated dev dependencies
- Updated prettier and linter configs
- Ran catalog generation script, since updating `csv-parse` seems to lead to the output being alphabetized


.eslintrc.json
```diff
@@ -57,7 +57,9 @@
     "jsdoc/require-returns-description": "error",
     "jsdoc/check-alignment": "error",
     "jsdoc/check-param-names": "error",
-    "react-hooks/exhaustive-deps": "error"
+    "react-hooks/exhaustive-deps": "error",
+    "@typescript-eslint/no-empty-object-type": "warn",
```
hunterckx, line 61: I think it might be worth leaving `@typescript-eslint/no-empty-object-type` at its default settings, and just removing the empty `Props` interface and its use as a type parameter in `figure.styles.tsx` -- as far as I can see there's no reason for that interface to exist, since the component is never passed any props anyway (though let me know if there's some particular reason to have this rule as a warning!)

hunterckx (changes requested):

.eslintrc.json
jpaten: Thanks for catching this @hunterckx!

hunterckx (approved):

Issue #206, opened 2025-01-09
Create YAML files for initial workflows and assemblies


NoopDog:

d-callan:
initial yaml has been delivered. we still have an outstanding task in #282 to improve the way we produce the workflows.yml, and automagically pick up input parameters etc

PR #207, opened 2025-01-09
feat: add workflow and assemblies yml files first pass (#206)

NoopDog:
Added files for Assembly (Input) and Workflows matching this [evolving schema doc](https://docs.google.com/spreadsheets/d/1j0fMLOSLK5NytI770LqtKXKVcDCV27OepSrQtAhyhQM/edit?gid=627506093#gid=627506093).

### DIFF
--- a/files/source/galaxy_workflows.yml
+++ b/files/source/galaxy_workflows.yml
@@ -1,0 +1,15 @@
+workflows:
+  - trs_id: "https://dockstore.org/api/ga4gh/trs/v2/tools/#workflow/github.com/iwc-workflows/rnaseq-pe/main/versions/v0.9"
+    type: "REGULATION"
+    description: "Workflows for the analysis of ChIP-seq, ATAC-Seq, and beyond."
+    ploidy: "any"
+
+  - trs_id: "https://dockstore.org/api/ga4gh/trs/v2/tools/#workflow/github.com/iwc-workflows/chipseq-pe/main/versions/v0.12"
+    type: "TRANSCRIPTOMICS"
+    description: "Analyze bulk or single-cell RNA seq data using a variety of approaches."
+    ploidy: "any"
+
+  - trs_id: "https://dockstore.org/api/ga4gh/trs/v2/tools/#workflow/github.com/iwc-workflows/haploid-variant-calling-wgs-pe/main/versions/v0.1"
+    type: "VARIANT_CALLING"
+    description: "Identify nucleotide polymorphisms and short indels from Illumina and Element data."
+    ploidy: "haploid"
--- a/files/source/galaxy_workflows.yml
+++ b/files/source/galaxy_workflows.yml
@@ -1,46 +1,40 @@
 workflows:
-  # this list of workflows and their annotations are a wip and will be updated as i learn more
-  # about their status and the tools used, etc.
-  - trs_id: "#workflow/github.com/iwc-workflows/chipseq-pe/main"
+  - trs_id: "https://dockstore.org/api/ga4gh/trs/v2/tools/#workflow/github.com/iwc-workflows/chipseq-pe/main/versions/v0.12"
     type: "REGULATION"
     description: "Workflows for the analysis of ChIP-seq, ATAC-Seq, and beyond."
     ploidy: "any"
 
-  - trs_id: "#workflow/github.com/iwc-workflows/chipseq-sr/main"
+  - trs_id: "https://dockstore.org/api/ga4gh/trs/v2/tools/#workflow/github.com/iwc-workflows/chipseq-sr/main/versions/v0.12"
     type: "REGULATION"
     description: "Workflows for the analysis of ChIP-seq, ATAC-Seq, and beyond."
     ploidy: "any"
 
-  - trs_id: "#workflow/github.com/iwc-workflows/rnaseq-pe/main"
+  - trs_id: "https://dockstore.org/api/ga4gh/trs/v2/tools/#workflow/github.com/iwc-workflows/rnaseq-pe/main/versions/v1.2"
     type: "TRANSCRIPTOMICS"
     description: "Analyze bulk or single-cell RNA seq data using a variety of approaches."
     ploidy: "any"
 
-  - trs_id: "#workflow/github.com/iwc-workflows/rnaseq-sr/main"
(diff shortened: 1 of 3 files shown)


files/source/galaxy_workflows.yml
```diff
@@ -0,0 +1,15 @@
+workflows:
+  - trs_id: "https://dockstore.org/api/ga4gh/trs/v2/tools/#workflow/github.com/iwc-workflows/rnaseq-pe/main/versions/v0.9"
+    type: "REGULATION"
+    description: "Workflows for the analysis of ChIP-seq, ATAC-Seq, and beyond."
```
d-callan, line 4: are these descriptions for the workflows? or for the 'type' ?

there will eventually be many workflows belonging to a 'type' (or whatever we end up calling it.. type is too vague imo), and i wonder if this source of duplication implies a need for a new yaml file?

NoopDog: Yes, reading the descriptions, I agree they should like they describe the workflow type rather than a specific workflow within the type. I think this came from having only one workflow per type for the initial kickoff. Sounds like we do need a type (or "modality?) tab.

NoopDog: I think we could also add (or generate) links to view the workflow details in GitHub, Dockstore etc. or wherever they live.

files/source/galaxy_workflows.yml
```diff
@@ -0,0 +1,15 @@
+workflows:
+  - trs_id: "https://dockstore.org/api/ga4gh/trs/v2/tools/#workflow/github.com/iwc-workflows/rnaseq-pe/main/versions/v0.9"
```
dannon, line 2: I think the actual trsID here is just `#workflow/github.com/iwc-workflows/rnaseq-pe/main`, though it does bring up how we want to pin specific versions (or not).  Do we want an explicit extra step to update this annotation as new versions are released, or do we want 'latest' @mvdbeek

files/source/galaxy_workflows.yml
dannon: Can/should we align these 'types' with the collection as organized on dockstore (and iwc site)?  (https://dockstore.org/organizations/iwc)

This can be programmatically fetched if so.

files/source/galaxy_workflows.yml
d-callan: the word 'latest' scared me :fearful:

files/source/galaxy_workflows.yml
```diff
@@ -0,0 +1,46 @@
...
+    type: "VARIANT_CALLING"
+    description: "Identify nucleotide polymorphisms and short indels from Illumina and Element data."
+    ploidy: "haploid"
+
+  # im starting to wonder looking at this if the default for ploidy shouldnt be any
+  # maybe it could be multivalued [haploid, diploid].. might not matter depending on ui
+  - trs_id: "#workflow/github.com/iwc-workflows/generic-variant-calling-wgs-pe/main"
+    type: "VARIANT_CALLING"
+    description: "Identify nucleotide polymorphisms and short indels from Illumina and Element data."
+    ploidy: "any"
```
mvdbeek, line 34: I think this is meant for haploid / non-diploid organisms. @nekrut I think you wrote this one ?

files/source/galaxy_workflows.yml
```diff
@@ -0,0 +1,46 @@
...
+
+  - trs_id: "#workflow/github.com/iwc-workflows/assembly-with-flye/main"
+    type: "ASSEMBLY"
+    description: "Best practices for assembly of prokaryotic and eukaryotic genomes sequenced with a variety of technologies."
+    ploidy: "any"
+
+  # idk yet what makes this specific to bacteria, or if it can be easily generalized
```
mvdbeek, line 41: Uses https://github.com/tseemann/shovill

files/source/galaxy_workflows.yml
```diff
@@ -0,0 +1,46 @@
...
+  # im starting to wonder looking at this if the default for ploidy shouldnt be any
+  # maybe it could be multivalued [haploid, diploid].. might not matter depending on ui
+  - trs_id: "#workflow/github.com/iwc-workflows/generic-variant-calling-wgs-pe/main"
+    type: "VARIANT_CALLING"
+    description: "Identify nucleotide polymorphisms and short indels from Illumina and Element data."
+    ploidy: "any"
+
+  - trs_id: "#workflow/github.com/iwc-workflows/assembly-with-flye/main"
+    type: "ASSEMBLY"
+    description: "Best practices for assembly of prokaryotic and eukaryotic genomes sequenced with a variety of technologies."
```
mvdbeek, line 38: The description is repeated, but also weird ? Maybe stick to the `Workflow for ...` pattern you've used above.

files/source/galaxy_workflows.yml
d-callan: Thanks. maybe it's enough to annotate as haploid then..

files/source/galaxy_workflows.yml
d-callan: Yea I think these could use work. For now I was just replicating what was on the site as category descriptions.

files/source/galaxy_workflows.yml
mvdbeek: > I think the actual trsID here is just `#workflow/github.com/iwc-workflows/rnaseq-pe/main`, though it does bring up how we want to pin specific versions (or not)

No, this is the full TRS id as we need it. No, no latest, we should always use pinned versions.

NoopDog:
Ok then, we have incorporated the assemblies.yml on to main. I am will close this PR and the related ticket and reopen two new tickets, one for workflows and one for the outbreak information.

d-callan:
@NoopDog ive just about got ready the assemblies yaml w the 778 we had before, now annotated and in the right format etc. let me know where to put that and when..

NoopDog:
@d-callan, you can PR your new assemblies in https://github.com/galaxyproject/brc-analytics/blob/main/catalog-build/source/assemblies.yml in a new PR against main.

NoopDog:
But @d-callan I hear from @nekrut we may want to focus on the current 21 for a bit?

d-callan:
idk when exactly we need them, just remembered hearing on the call w wiriya it sounded like we wanted them back soon. i dont mind holding on to it a bit longer, or making a draft PR and letting it sit there if others would like a look.

NoopDog:
Cool, I will set you up for the draft PR in a bit. @d-callan Thx!

Issue #208, opened 2025-01-14
BRC v0 Release


NoopDog:
1. Rename the current "Datasets" tab on the header to "Organisms"
2. Restrict organisms to the ones in the phase1_genomes list.
3. Read the assembly metadata from NCBI
4. Selecting the organism name shows the organism detail page on the /organism/id route
5. On this route we list the assemblies (we currently have one per) on an "Assemblies" tab.
6. Add strain to the assemblies list 
7. Show the View and Analyze buttons on the "Assemblies" List

hunterckx:
The process for building the input data (as done in #178) is:
- Get the list of organisms from the google sheet
- Get the taxonomy IDs from NCBI based on the organism names (the spreadsheet does specify taxonomy IDs so I wonder if this step is unnecessary)
- Output the organism list, consisting of: name from NCBI; taxonomy ID from NCBI; accession from the spreadsheet (there's assumed to only be one); correspondingly, assembly count 1; and tags from the spreadsheet (and also anything else that's in the spreadsheet but I don't think we use any of the other columns)
- Get genomes from NCBI for each organism (by taxonomy ID), filtering by whether the accession matches the one given in the spreadsheet
- Combine with assembly list JSON to add UCSC Browser URL
- Add gene model URLs scraped from UCSC based on accessions (I didn't write this part so I don't know the exact details of what's going on)
- Output genome list

NoopDog:
Complete

PR #209, opened 2025-01-16
feat: build data from ncbi and add organisms with assembly lists (#208)

hunterckx:


NoopDog (approved):
Awesome, thx @hunterckx

PR #223, opened 2025-01-21
feat: update titles on assembly detail page (#221)

### LINKED ISSUE #221: Update Assembly Detail page header to highlight the assembly id
Image

Currently, the assembly detail page calls out Choose analysis methods" three times but the assembly id is only in the URL and the right-hand detail.

Let's update the header/breadcrumbs to:

1. Have the breadcrumbs say Assemblies/{org name]/[assemblyid}
2. Have the h1 say {assemblyid)
3. Update the Select Analysis Method to "Select Analysis Workflow"

Image

hunterckx:
Closes #221

I've also switched the page's heading to use the same component (`BackPageHero`) as the organism detail


NoopDog (approved):
Nice! Thx @hunterckx

PR #229, opened 2025-01-23
refactor: get accessions from yaml file when building catalog source (#228)

### LINKED ISSUE #228: Read assemblies from YAML file and convert to standard directory names
#### Need 
Currently the assemblies are read from the Phase one genomes spreadsheet.

We need instead to read them from a file under GitHub control.

#### Approach

1. Rename "files" to "catalog-build"
2. Move "out" directory to "catalog" at the same level as catalog-build
3. In the "catalog-build/source" directory remove any unused sheets
4. in "catalog-build/source" add in an assemblies.yaml file that looks like [this](https://github.com/galaxyproject/brc-analytics/pull/207/files) but without the ploidy slot and keeping the scientific name in the comments
5. The assemblies should be taken from the current "phase one genomes" sheet and this should match the most recent build.
6. Use the new assemblies.yaml file to be the source of the  assemblies in the build scripts.
7. Make the build deposit the new assembly.json and organoid.json files in the new "catalog" directory.

Note this is similar to the HPRC directory layout on main there.

hunterckx:
Closes #228


NoopDog (approved):
Super, thanks @hunterckx

PR #258, opened 2025-02-05
feat: create catalog build python package (#251)

### LINKED ISSUE #251: Local Python package for data build scripts

hunterckx:
Closes #251

~~Drafted for now in case further updates are needed to support galaxyproject/ga2#23~~


catalog-build/package/catalog_build/build.py
```diff
@@ -0,0 +1,160 @@
...
+
+  report_missing_values_from("accessions", "matched in assembly list", genomes_with_species_df["accession"], assemblies_df["genBank"], assemblies_df["refSeq"])
+  
+  genomes_df = gen_bank_merge_df.combine_first(ref_seq_merge_df)
+
+  if do_gene_model_urls:
+    genomes_df = add_gene_model_url(genomes_df)
+    report_missing_values("accessions", "matched with gene model URLs", genomes_df["accession"], genomes_df["geneModelUrl"].astype(bool))
+  else:
+    genomes_df["geneModelUrl"] = ""
```
hunterckx, line 156: I've added an option to disable gene model URLs, because the process of getting them can take some time which makes it harder to test changes to the catalog build script -- not certain this will actually come in handy in the future but there were a couple times here where it would've been useful so I figure it's a nice thing to have available

catalog-build/package/catalog_build/build.py
```diff
@@ -0,0 +1,160 @@
...
+  for series in partial_values_series:
+    present_values_mask |= all_values_series.isin(series)
+  report_missing_values(values_name, message_predicate, all_values_series, present_values_mask)
+
+def report_missing_values(values_name, message_predicate, values_series, present_values_mask):
+  missing_values = values_series[~present_values_mask]
+  if len(missing_values) > 0:
+    if len(missing_values) > len(values_series)/2:
+      present_values = values_series[present_values_mask]
+      print(f"Only {len(present_values)} of {len(values_series)} {values_name} {message_predicate}: {", ".join(present_values)}")
```
hunterckx, line 124: If there are more missing values than present values, the present values are reported instead, which makes the message more meaningful in extreme cases

NoopDog (approved):
Thx @hunterckx

Issue #260, opened 2025-02-05
Take workflows from yml config


NoopDog:
The workflows are currently "hard coded". We need to take them from a config file. We also need to identify any unique ways the query is built for the workflows and add those to the workflow config if possible so we know how to map the context (one or more selected assemblies and the assembly type, etc. to the workflow API call.)

The current best effort for the workflow config is here:

https://github.com/galaxyproject/brc-analytics/blob/11366c6d9b2a47cfef397507db21015ef295b3ea/files/source/galaxy_workflows.yml

Lets exclude the two ASSEMBLY workflows we can add those in as a separate ticket.

The workflow initial data model is here https://docs.google.com/spreadsheets/d/1j0fMLOSLK5NytI770LqtKXKVcDCV27OepSrQtAhyhQM/edit?gid=0#gid=0

From there we will have a yml file for:

1. workflow_categories.yml
2. workflows.yml

Processing these, we will generate a workflows.json that we can use in the application at build time to configure the workflows.

#### Workflow category vs Workflow

Our workflow categories are what we display now on the site

<img width="824" alt="Image" src="https://github.com/user-attachments/assets/38eccff2-343f-47eb-8108-d0e9be12e6d4" />

Currently, we support only one workflow per category, and we do not yet show the workflow name and workflow description. Soon we will show similar to this:

<img width="493" alt="Image" src="https://github.com/user-attachments/assets/7e1d6737-7e43-4712-b4de-062c4f1b34d8" />

Where the workflow name and descriptions are in the drop down.

You can also see workflow names and descriptions on the https://iwc.galaxyproject.org/ site.
e.g.

<img width="966" alt="Image" src="https://github.com/user-attachments/assets/bb474c72-b0f8-4e41-9e3f-95e83e43b644" />

MillenniumFalconMechanic:
Completed.

PR #265, opened 2025-02-08
feat: derive workflows from yaml files (#260)

hunterckx:
I've made it so only the TRS IDs we were using before are available, but that that means that transcriptomics is no longer available, because the ID in the YAML is for v1.2 rather than the v0.9 that we had before


app/utils/galaxy-api.ts
```diff
@@ -71,11 +73,11 @@ function buildFastaUrl(identifier: string): string {
  * @returns `request_state` value for the workflow landings request body.
  */
 function getWorkflowLandingsRequestState(
-  workflowId: WORKFLOW_ID,
+  workflowId: string,
   referenceGenome: string,
   geneModelUrl: string | null
 ): WorkflowLandingsBodyRequestState {
-  if (workflowId === WORKFLOW_ID.VARIANT_CALLING && geneModelUrl) {
+  if (workflowId === WORKFLOW_ID_VARIANT_CALLING && geneModelUrl) {
```
hunterckx, line 80: The workflow landings request body is still varied based on a hardcoded TRS ID

NoopDog:
Oh you can update the yml @hunterckx

hunterckx:
@NoopDog Updated

NoopDog (approved):
LGTM. Thanks @hunterckx

PR #283, opened 2025-02-12
feat: add linkml schemas for data source files (#269, #270)

### LINKED ISSUE #269: Create JSON schema for assembly.yml and use for validating the file in vscode

### LINKED ISSUE #270: Create a json schema for the workflows.yml and use it to validate the file in vscode

hunterckx:
Closes #269
Closes #270

EDIT: Some additional notes:
- The descriptions are pretty minimal
- I've add checks to the GitHub workflow for:
  - Linting the schemas
  - Converting the schemas to Python, which from my experiments is the most reliable way I've found to check for errors
  - Validating the catalog data according to the schemas

Also updates ploidy values to be all caps, for consistency

Questions (EDIT: now resolved):
- What URIs should we use for the schema IDs? I've used URLs to brc-analytics.org, but another option might be to use GitHub URLs; either way we would ideally want the schemas to actually exist at the URLs, which would require some slight extra work in the former case
- How should the JSON schemas be incorporated? Do we want some subfolder containing them?


catalog/schema/assemblies.yaml
```diff
@@ -0,0 +1,28 @@
+id: https://github.com/galaxyproject/brc-analytics/blob/main/catalog/schema/assemblies.yaml#
```
hunterckx, line 1: I've put `#` on the ends of the IDs, so that derived IDs will still be URLs that actually refer to something (of course currently they don't refer to anything but they should once this is merged)

catalog/schema/workflows.yaml
```diff
@@ -0,0 +1,46 @@
...
+        range: Workflow
+
+  Workflow:
+    description: A workflow.
+    attributes:
+      trs_id:
+        description: The workflow's TRS ID.
+        required: true
+        range: string
+      type:
```
mvdbeek, line 31: Could this be called categories and be an array since some workflows fit in more than one category ? I'd reserve type for iterations of the workflow format itself (we currently use .ga files but plan to use gxformat2 in the future).

catalog/schema/enums/workflow_ploidy.yaml
```diff
@@ -0,0 +1,9 @@
+id: https://github.com/galaxyproject/brc-analytics/blob/main/catalog/schema/enums/workflow_ploidy.yaml#
+name: enums_workflow_ploidy
+
+enums:
+  WorkflowPloidy:
+    description: Possible ploidies supported by workflows.
+    permissible_values:
+      ANY:
+      HAPLOID:
```
mvdbeek, line 9:
```suggestion
      HAPLOID:
      DIPLOID:
```

mvdbeek (commented):
Thanks, that looks good!

catalog/schema/enums/workflow_type.yaml
```diff
@@ -0,0 +1,14 @@
...
+enums:
+  WorkflowType:
+    description: Set of IDs of workflow categories.
+    permissible_values:
+      VARIANT_CALLING:
+      TRANSCRIPTOMICS:
+      REGULATION:
+      ASSEMBLY:
+      GENOME_COMPARISONS:
+      PROTEIN_FOLDING:
```
mvdbeek, line 14: It might make sense to add a placeholder like `OTHER` so we can generate a valid catalog if a category doesn't match yet ?

mvdbeek:
I was thinking maybe we could settle on something like this ?
```yaml
workflows:
- trs_id: '#workflow/github.com/iwc-workflows/haploid-variant-calling-wgs-pe/main/versions/v0.1'
  workflow_categories:
  - VARIANT_CALLING
  workflow_name: Paired end variant calling in haploid system
  workflow_description: Workflow for variant analysis against a reference genome in
    GenBank format
  ploidy: HAPLOID
  parameters:
    Annotation GTF:
      class: File
      ext: gtf.gz
      url: {{ gene_model_url }}
    Genome fasta:
      class: File
      ext: fasta.gz
      url: {{ assembly_fasta_url }}
  active: false
```

If I try to automate this for all workflows within the currently defined BRC categories I am generating this: https://gist.github.com/mvdbeek/b53d73a5fcba2677b0706bcd7ab10704

Importantly, it would be relatively easy to keep this updated (and the parameters in sync) whenever we merge a new workflow.

NoopDog:
Thanks, @mvdbeek, for the review. We might accept the small changes and open tickets for the larger ones to stay focused! 
Stay tuned!

NoopDog (approved):
Thanks @hunterckx

PR #290, opened 2025-02-13
feat: landing page organism viz - data (#289)

### LINKED ISSUE #289: Visualization for heirarchical display of genomes for homepage
Develop a visually engaging and functional hierarchical view of available genomes for the homepage. This visualization should make it easy for users to quickly browse and explore the genomic data available at brc-analytics.org while also providing eye candy.

d-callan:
OLD:
working our way towards a viz for the landing page. my understanding is that this pr should do the following:
 - [ ] provide a summary of what organisms/ assemblies we currently support as a sunburst
 - [ ]  for segments of the sunburst w ncbi tax ids that match our 'taxonomic group' filter, allow a user to select that segment and go to a filtered organism/ assembly page
 - [ ] be pretty lol
 - [x] automagically produce the required data via ncbi api calls

resolves #289 

UPDATED: 
tbh i no longer know what issue if any this should be tied to, as its been broken into three (pending) PRs now. this one represents just the data. the actual sunburst is somewhere else now..


dannon:
Here's the current view of it as demo'd in the call today;

image

dannon:
Note to myself that the 'happy accident' of parent nodes including themselves in the count, causing the organic-looking decreasing arc size is visually interesting but I should fix the count.

hunterckx (changes requested):
@d-callan I've noticed a few potential issues here:
- It looks like maybe the tree JSON hasn't been updated since the change to the taxonomic ranks? It still has superkingdoms in it
- I wasn't able to run the script, because the URL for the `filtered_subtree` API is too long
  - I notice now that NCBI has APIs such as `/taxonomy/filtered_subtree`, which takes the list of IDs in the body of a POST request rather than in the URL -- maybe that would be the easiest solution here? ([Documentation here](https://www.ncbi.nlm.nih.gov/datasets/docs/v2/api/rest-api/#post-/taxonomy/filtered_subtree))
- I'm not sure this actually matters as it is, but it might be best to make the taxonomy IDs unique within the list before passing them to any APIs, because the `taxonomyId` column from the dataframe may have multiple of the same ID
- Please update `catalog/build/py/package/setup.py` to increase the version number to `2.0.0` (as this update includes a breaking change to how `build_files` is called)

Thanks!

mvdbeek (approved):
Thank you @dannon @d-callan !

dannon:
This chunk should be good to go -- I've rebased it again, followed up on all but one of the suggestions (uniqing the tree handling) because I want to handle that in a separate PR where I also make the tree sorting stable.

d-callan:
Did we also manage to make sure both the domains and realms are under the root together? Just so I know its status when I'm back mon..

catalog/build/py/package/catalog_build/build.py
```diff
@@ -127,15 +129,175 @@ def get_species_row(taxon_info, taxonomic_group_sets, taxonomic_levels):
...
+      else:
+        print(f"rank not found for tax_id: {tax_id}")
+
+
+def ncbi_tree_to_nested_tree(node_id, edges, taxonomy_ids):
+  children = edges.get(str(node_id), {}).get("visible_children", [])
+  children = [str(num) for num in children]
+  # ncbi results odd again, dup children
+  children = set(children)
+  if (len(children) > 0 or node_id in taxonomy_ids):
```
hunterckx, line 284:
```suggestion
  if (len(children) > 0 or int(node_id) in taxonomy_ids):
```

`taxonomy_ids` contains ints

catalog/build/py/package/catalog_build/build.py
```diff
@@ -471,10 +639,18 @@ def build_files(
...
   
   if qc_report_path is not None:
     qc_report_text = make_qc_report(**qc_report_params)
     with open(qc_report_path, "w") as file:
       file.write(qc_report_text)
+
+  if len(taxonomic_levels_for_tree) > 0:
+    # Use the taxonomy IDs from the genomes_df to build the species tree
+    # Pass the previously fetched species_info to avoid another API call
+    species_tree = get_species_tree(genomes_df["taxonomyId"], taxonomic_levels_for_tree, species_info)
```
hunterckx, line 652:
```suggestion
    species_tree = get_species_tree(list(genomes_df["taxonomyId"]), taxonomic_levels_for_tree, species_info)
```

Keeping the taxonomy IDs as a Pandas series causes issues with the `in` operator down the line

hunterckx (changes requested):
Thanks @dannon! I did notice one more thing, which is that some species seem to be missing from the tree -- I think the missing ones are ones that have no children, and these suggested edits plus a rebuild of the tree should fix the issue

hunterckx (approved):
Awesome, looks good to me! Thanks all

PR #295, opened 2025-02-13
feat: fetch raw/primary data information from sra

Smeds:
Added the possibility of fetch primary data from SRA and output a new tsv file. Will be used by GenomeArk 2 and is optional to run.


hunterckx (changes requested):
- Since primary data is only saved when `extract_primary_data` is true, could `primary_output_path` be made optional, and the changes to `build-files-from-ncbi.py` reverted?
- The new dependencies should be added to `requirements.txt` (although the branch should be rebased first, since there have been changes to the requirements on main)

Thanks @Smeds!

Smeds:
> * Since primary data is only saved when `extract_primary_data` is true, could `primary_output_path` be made optional, and the changes to `build-files-from-ncbi.py` reverted?
> 
>     * The new dependencies should be added to `requirements.txt` (although the branch should be rebased first, since there have been changes to the requirements on main)
> 
> 
> Thanks @Smeds!

Done! I have modified the code according to your comments

hunterckx (approved):

Issue #297, opened 2025-02-14
Create organism.yaml file to annotate organism with manually curated information for workflow matching


NoopDog:
From our initial spreadsheet data model...
https://docs.google.com/spreadsheets/d/1j0fMLOSLK5NytI770LqtKXKVcDCV27OepSrQtAhyhQM/edit?gid=1942006530#gid=1942006530

1. Create a schema for an organism list and  organism with:
   1. tax_id
   2. ploidy -  where ploidy comes from the enum we already have.

2. create an initial organism.yml with the  current tax_ids for our 21 organisms (you can download a list from the site!)
3. Create a draft PR and we will ask for ploidy to be populated in the organism.yml
4. Add ploidy into the organism.json during the build data process matching from the organism.yml
5. On the assembly workflow page, remove the existing filters, allowing only the original 3 workflows.
6. Instead, filter out any workflows where workflow ploidy is not compatible with the organism ploidy.

NoopDog:
Complete

PR #299, opened 2025-02-15
feat: only show workflows with compatible ploidies (#297)

hunterckx:
Notes:
- As part of this, I've made it so the TypeScript types used for the source data are derived from the LinkML schemas
- Despite being named `organisms.yml`, the new source file never interacts with the organism list used in the catalog, since ploidy is only needed at the assembly level

To address:
- [x] Can we assume that every assembly for a given species will have the same ploidy? (I've set it up under that assumption since that's what's implied by the ticket)
- [x] Should it be allowed for assemblies to end up with no ploidy assigned to them (current behavior), or should the catalog build script throw an error or something if none is found?


NoopDog:
Poloidy is required options are to stop the build or skip the assembly and continue.

I guess the latter. 

E.g warning no ploidy found for taxid xxx skipping assembly yyy.

NoopDog:
Only one ploidy per taxid. We can add the description found in the spreadsheet "ploidy at the usually sampled life stage" check the sheet for the quote.

NoopDog:
Hi @hunterckx 

 I think what we need to do here is:

1. Always show the "Coming Soon" workflow categories.
2. If a workflow category is empty because of the filters, label it with the "Coming Soon" tag and make it not clickable, just like the other "Coming Soon" workflows.

hunterckx:
@NoopDog Updated!

NoopDog (approved):
Thanks @hunterckx

PR #306, opened 2025-02-18
feat: add filters for taxonomic levels to organisms list (#301)

### LINKED ISSUE #301: Add filters for taxonomic levels on the organism page
#### Need
The s[unburst vis organism selecto](https://github.com/galaxyproject/brc-analytics/pull/290)r will have a final step where users can move to the organism list with the given filter applied. To support this we will need to enable filtering by the various taxonomic levels.

"superkingdom",
  "kingdom",
  "phylum",
  "class",
  "order",
  "family",
  "genus",
  "species",
  "strain",

Lets add these filters on the organisms page under a "Taxonomy Level" heading. 
There may be some code in the linked PR above that helps get the taxonomic levels per taxid with the associated level. 

Will add this to the assemblies page as well on a different ticket.  Lets get the organisms going the way we want and then continue.

hunterckx:
Closes #301

- The app is updated to use the new `taxonomicLevelSpecies` field instead of the old `species`, but assemblies are unchanged and `species` is left in the package for compatibility.
- I've included an "other" category, because the tree in #290 (EDIT: as of writing anyway) includes taxonomic levels other than the ones specified (the tree API is supposed to only return specified levels, but I think maybe it always includes the requested taxon as a child node, so when we request a non-strain infraspecific taxon that gets included despite not being one of the requested levels).
- Not sure what assumptions we should be making about whether taxa at different levels are available -- I've typed everything except species as `string | null`, but assumed species will always be there and typed it as `string`. Kingdom, strain, and other infraspecific taxa can definitely be absent.


catalog/build/ts/build-catalog.ts
```diff
@@ -155,14 +173,22 @@ function buildWorkflow(
...
-  return parseCsv(content, {
+  const result = parseCsv(content, {
     columns: true,
     delimiter,
     relax_quotes: true,
   });
+  if (checkKeys && result[0]) {
+    for (const key of checkKeys) {
+      if (!Object.hasOwn(result[0], key))
+        throw new Error(`Missing column ${JSON.stringify(key)} in ${filePath}`);
```
hunterckx, line 188: Added this in case the taxonomic levels used in the Python get out of sync with the ones used in the TypeScript -- maybe it would be nicer to have them both draw from shared definitions somehow but I'm not sure what the best way to do that would be (I suppose the ideal might be LinkML checks in the Python so maybe this is fine until we get that)

NoopDog (approved):

Issue #322, opened 2025-02-24
Review and prioritize additional workflows to be added to the BRC Site


NoopDog:
#### Need

Currently, the BRC site has three workflows, but the IWS has 100 or so. We have a prioritized list of the [first eight workflows](https://github.com/galaxyproject/brc-analytics/blob/11366c6d9b2a47cfef397507db21015ef295b3ea/files/source/galaxy_workflows.yml) to add to BRC; however, we need to understand if these workflows require any additional parameters to be sent.

The initial eight are (tho the structure has changd for the YAML)

```
workflows:
  - trs_id: "https://dockstore.org/api/ga4gh/trs/v2/tools/#workflow/github.com/iwc-workflows/chipseq-pe/main/versions/v0.12"
    type: "REGULATION"
    description: "Workflows for the analysis of ChIP-seq, ATAC-Seq, and beyond."
    ploidy: "any"

  - trs_id: "https://dockstore.org/api/ga4gh/trs/v2/tools/#workflow/github.com/iwc-workflows/chipseq-sr/main/versions/v0.12"
    type: "REGULATION"
    description: "Workflows for the analysis of ChIP-seq, ATAC-Seq, and beyond."
    ploidy: "any"

  - trs_id: "https://dockstore.org/api/ga4gh/trs/v2/tools/#workflow/github.com/iwc-workflows/rnaseq-pe/main/versions/v1.2"
    type: "TRANSCRIPTOMICS"
    description: "Analyze bulk or single-cell RNA seq data using a variety of approaches."
    ploidy: "any"

  - trs_id: "https://dockstore.org/api/ga4gh/trs/v2/tools/#workflow/github.com/iwc-workflows/rnaseq-sr/main/versions/v1.2"
    type: "TRANSCRIPTOMICS"
    description: "Analyze bulk or single-cell RNA seq data using a variety of approaches."
    ploidy: "any"

  - trs_id: "https://dockstore.org/api/ga4gh/trs/v2/tools/#workflow/github.com/iwc-workflows/haploid-variant-calling-wgs-pe/main/versions/v0.1"
    type: "VARIANT_CALLING"
    description: "Identify nucleotide polymorphisms and short indels from Illumina and Element data."
    ploidy: "haploid"

  - trs_id: "https://dockstore.org/api/ga4gh/trs/v2/tools/#workflow/github.com/iwc-workflows/generic-variant-calling-wgs-pe/main/versions/v0.1.1"
[...]

PR #335, opened 2025-02-25
feat: added additional workflows (#322)

d-callan:
changed strategy a bit to try to go around the issue where some workflows may need support for additional pre-populated parameters, but still get some fun things on the site quickly.

this should get single read options for rnaseq and chipseq, add cutandrun, atacseq and some scrnaseq stuff.

also improved display names of some things. perhaps we want to improve those in iwc in future, and if so, probably an issue or two need filed somewhere for that?

feedback welcome


mvdbeek:
Can you please update this in the IWC? If we diverge here it will be hard to keep this updated.

NoopDog:
Hi @mvdbeek I was hoping we could move to taking this info directly from IWC. I will create a ticket for that in a bit.  I'm hoping we don't need to block this PR on the IWC update.

NoopDog:
Hi @mvdbeek. Would you be able to review this for technical accuracy so we can merge it?  We definatly need a plan to avoid the duplication of the descriptions, but for right now. Do these look ok? Thanks!

mvdbeek:
You can merge this, but we're going to override this later.

hunterckx (approved):

Issue #323, opened 2025-02-24
Add back all assemblies


MillenniumFalconMechanic:
Add assemblies to [assemblies.yml](https://github.com/galaxyproject/brc-analytics/blob/main/catalog/source/assemblies.yml).
Make sure all ploidy is set for all organisms in [organisms.yml](https://github.com/galaxyproject/brc-analytics/blob/main/catalog/source/organisms.yml)

NoopDog:
From Element

<img width="570" alt="Image" src="https://github.com/user-attachments/assets/5358b86d-05e7-4e74-9bc9-8ea83c815a86" />

NoopDog:
We will make ploidy multi-valued on organism.yml to not over-restrict the workflows when the ploidy can vary over the lifecycle or for other reasons.

https://github.com/galaxyproject/brc-analytics/issues/334

PR #340, opened 2025-02-28
feat: add back all assemblies (#323)

d-callan:
holy cannoli batman

I have been on a journey, and i have learned some things:

1. it would be very good to be able to add or change 'taxonomic groups' without having to re-fetch all the data from ncbi api
2. the technical process of adding ploidy, or probably any other annotation on organisms needs some improvement. its too cumbersome currently, but i dont (yet!) have a good suggestion. ill think on it some more...
3. in the end i had maybe 30 things i couldnt figure out how to annotate for ploidy, and i just put all three haploid, diploid and polyploid
4. there are a handful of things i didnt add a 'taxonomic group' for, but i think for now at least its ok. i think i got the big ones, and things like adding 'Mollusca' or something on account of a single snail in vectorbase seemed slightly silly and probably overkill.
5. some number of these ploidy annotations are going to end up being wrong, just bc im human, there were soooooo many of them and the fungi are absurdly complicated


d-callan:
> as currently it's "Vertebra"

this actually made me laugh aloud. signs of delirium settling in w this particular task :joy:

hunterckx (approved):
Thanks @d-callan, looks good to me!

NoopDog (approved):

Issue #339, opened 2025-02-27
``parameters`` & ``active`` fields in workflow schema


mvdbeek:
So we can define parameters in workflows.yml.
xref https://github.com/galaxyproject/brc-analytics/pull/283#issuecomment-2657159212.

The active field is not entirely necessary, but would make it pretty easy to automate PRs that would add or update new workflows. When processing the source file you could just drop anything that is inactive. Otherwise one would need to edit down PRs to include only the new workflows, which would be a little more work than necessary considering we're probably not going to add all of the IWC workflows.

MillenniumFalconMechanic:
See also #351.

Issue #277, opened 2025-02-11
Can workflow TRS IDs point to latest?


NoopDog:

mvdbeek:
No, parameters can change and "latest" is a problem for reproducibility.

NoopDog:
OK then. @nekrut reply is as you expected ;)

PR #342, opened 2025-03-03
feat: update workflows from IWC manifest and set assembly-specific parameters (#351, #339)

mvdbeek:
Closes https://github.com/galaxyproject/brc-analytics/issues/339.

Run
```shell
python catalog/source/iwc_manifest_to_workflow_yaml.py --exclude-other
npm run build-brc-db
```
to fetch a list of current workflows from https://iwc.galaxyproject.org/workflow_manifest.json.
Only workflows for currently enabled categories are fetched.
If necessary, annotate parameters that require a reference genome id, fasta or gtf file with
`{{ assembly_id }}`, `{{ assembly_fasta_url }}`, `{{ gene_model_url }}`.
These values will be substituted with assembly-specific values at runtime.

### DIFF
--- a/catalog/build/py/iwc_manifest_to_workflows_yaml.py
+++ b/catalog/build/py/iwc_manifest_to_workflows_yaml.py
@@ -17,6 +17,7 @@ DOCKSTORE_COLLECTION_TO_CATEGORY = {
     "Epigenetics": WorkflowCategoryId.REGULATION,
     "Genome assembly": WorkflowCategoryId.ASSEMBLY,
 }
+MANIFEST_SOURCE_OF_TRUTH = ("trs_id", "workflow_name", "categories", "workflow_description")
 
 
 def read_existing_yaml():
@@ -99,10 +100,12 @@ def merge_into_existing():
     for versionless_trs_id, current_workflow_input in current.items():
         existing_workflow_input = existing.get(versionless_trs_id)
         if existing_workflow_input:
-            # someone has annotated ploidy and parameters, keep these, but also check
-            # that existing parameters still exist
-            current_workflow_input.ploidy = existing_workflow_input.ploidy
-            current_workflow_input.active = existing_workflow_input.active
+            # we'll keep whatever has been specified in the brc repo,
+            # and only update values that are in the iwc manifest
+            exisiting_dict = existing_workflow_input.model_dump()
+            new_dict = current_workflow_input.model_dump()
+            for key in MANIFEST_SOURCE_OF_TRUTH:
+                exisiting_dict[key] = new_dict[key]
             # check that specified parameters still exist
             current_workflow_parameter_keys = {param.key for param in current_workflow_input.parameters}
             for param in existing_workflow_input.parameters:
@@ -111,8 +114,7 @@ def merge_into_existing():
                     raise Exception(
                         f"{param.key} specified but is not part of updated workflow {current_workflow_input.trs_id}! Review and fix manually"
                     )
-            # keep existing parameters
-            current_workflow_input.parameters = existing_workflow_input.parameters
+            current_workflow_input = Workflow(**exisiting_dict)
         merged[versionless_trs_id] = current_workflow_input
     return merged
 


catalog/output/workflows.json
```diff
@@ -34,19 +34,19 @@
           "gtf": "{{ gene_model_url }}"
         },
         "ploidy": "ANY",
-        "trsId": "#workflow/github.com/iwc-workflows/fastq-to-matrix-10x/scrna-seq-fastq-to-matrix-10x-cellplex/versions/v0.6",
-        "workflowDescription": "This workflow processes the CMO fastqs with CITE-seq-Count and include the translation step required for cellPlex processing. In parallel it processes the Gene Expresion fastqs with STARsolo, filter cells with DropletUtils and reformat all outputs to be easily used by the function 'Read10X' from Seurat.",
-        "workflowName": "scRNA-seq_preprocessing_10X_cellPlex"
+        "trsId": "#workflow/github.com/iwc-workflows/fastq-to-matrix-10x/scrna-seq-fastq-to-matrix-10x-cellplex/versions/v0.6.2",
```
mvdbeek, line 37: This is what a normal update will look like, since the current diff for the whole PR is hard to read.

catalog/schema/workflows.yaml
```diff
@@ -45,3 +49,11 @@ classes:
         description: The description of the workflow.
         required: true
         range: string
+      parameters:
+        description: The parameters of the workflow. Dictionary with arbitrary depth, string keys and primitive values.
+        required: false
+        range: Any
```
d-callan, line 55: idk linkml almost at all, but im finding it surprising there isnt a way to do what we want here

catalog/source/iwc_manifest_to_workflows_yaml.py
```diff
@@ -0,0 +1,180 @@
...
+    workflow_description: str
+    ploidy: str
+    # readme: str
+    parameters: Optional[Dict[str, Any]] = None
+    active: bool = False
+
+    def as_dict(self):
+        # convert to brc category enum
+        d = asdict(self)
+        d["categories"] = sorted(
```
d-callan, line 44: so this means a workflow can have multiple categories? is that a thing we want?

catalog/source/iwc_manifest_to_workflows_yaml.py
```diff
@@ -0,0 +1,180 @@
...
+        for workflow in sorted_workflows:
+            if "OTHER" in workflow["categories"]:
+                workflow["categories"].remove("OTHER")
+                if not workflow["categories"]:
+                    print(f'Excluding workflow {workflow["trs_id"]}, category unknown')
+                    continue
+            final_workflows.append(workflow)
+        sorted_workflows = final_workflows
+    final_workflows = sorted_workflows
+    with open(WORKFLOWS_PATH, "w") as out:
```
d-callan, line 159: so if im understanding so far, the new flow is to overwrite the yml file w this script, then run the existing script to make workflows.json from the now semi-manually, semi-programmatically created yaml? i wonder if a better flow would be something like to put a list of only the workflows we want in workflows yml w only the fields that need manual curation (ploidy, soon to be tax id), and then add the params, descriptions, names, etc as part of the script that produces the json?

catalog/schema/generated/schema.ts
```diff
@@ -115,6 +120,10 @@ export interface Workflow {
     workflow_name: string,
     /** The description of the workflow. */
     workflow_description: string,
+    /** The parameters of the workflow. Dictionary with arbitrary depth, string keys and primitive values. */
+    parameters?: Any,
+    /** Determines if workflow should be included. */
+    active: boolean,
```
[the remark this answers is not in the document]
d-callan: ok. i see the description of the thing now. im not 100% convinced we need this field, but if we decide to keep it maybe we can think of a more verbose/ descriptive thing to call it?

catalog/source/iwc_manifest_to_workflows_yaml.py
```diff
@@ -0,0 +1,180 @@
...
+CATEGORIES = Literal[
+    "VARIANT_CALLING",
+    "TRANSCRIPTOMICS",
+    "REGULATION",
+    "ASSEMBLY",
+    "GENOME_COMPARISONS",
+    "PROTEIN_FOLDING",
+    "OTHER",
+]
+DOCKSTORE_COLLECTION_TO_CATEGORY = {
```
d-callan, line 22: can we/ do we verify somewhere that all the categories listed in the categories yml have a mapping? or do we want to try to do that is i guess the first q..

catalog/source/iwc_manifest_to_workflows_yaml.py
d-callan: also, is this automatically picking up new versions of workflows? is that a thing we want?

NoopDog:
@mvdbeek, we are looking at this in detail. In a few days, we will have a PR into this PR for you to review. Thanks for putting this together.

NoopDog:
I have converted this to a draft PR while we make suggestions.

hunterckx:
Hey @mvdbeek, I'm trying think of what we could do to make the model more specific and avoid using `Any` for `parameters`, but I think I could do with a bit of background on what's required of the parameter definitions.

It looks like the parameter definitions currently come in a pretty limited number of forms (namely strings to be replaced with a variable and a few variations on dicts with a `class` field) -- is there a reason we would want to keep it open to arbitrary dictionaries?

I'm also curious what the intended use of those parameters with a `class` is, as they don't appear to be used currently.

mvdbeek:
> is there a reason we would want to keep it open to arbitrary dictionaries?

yes, that's how you submit arrays or things, or set of nested parameters. Note how the parameter names are entirely configurable in a workflow, so I don't think a fixed schema will do.

> I'm also curious what the intended use of those parameters with a class is, as they don't appear to be used currently.

These are in the source only, so you can see at a glance what the type of the parameter is and you don't need to go checking in the workflow whether we need a assembly id or a fasta file. It's a convenience thing that makes setting parameters to assembly specific inputs easy.

catalog/source/iwc_manifest_to_workflows_yaml.py
mvdbeek: I don't see what the advantage is with your suggestion. As a maintainer you look at a single file, set what's needed to be set and you toggle the active flag. Updates will work fine for the most part unless a specified parameter has changed, in which case the script fails and you have to fix this manually.

Yes, we want the latest versions, this is a safe way to do what https://github.com/galaxyproject/brc-analytics/issues/277 was aiming at.

catalog/source/iwc_manifest_to_workflows_yaml.py
mvdbeek: Put it in a file and read it from there ? I don't think this needs to be part of this PR.

hunterckx:
Thanks for the clarification @mvdbeek! I've opened pull request mvdbeek#1 targeting this branch with my attempt at refining the model -- see additional notes there

(cc @NoopDog)

d-callan:
who is reviewing this, me?

README.md
```diff
@@ -58,6 +58,21 @@ the `# Anopheles gambiae` comment:
 - accession: XXX_000000000.0
 ```
 
+### Updating workflows
+
+Run
+
+```shell
+python catalog/source/iwc_manifest_to_workflow_yaml.py --exclude-other
```
d-callan, line 66: looks like we have a new helper in package.json, but readme isnt updated to reflect this?

d-callan (approved):
ah well, i went and reviewed it

i still dont prefer this flow, where a thing partly programmatically produced is in 'source' where it can be (and is meant to be) modified after. but i wont stand in its way for that

hunterckx (changes requested):
I'd like to request a couple changes:
- Looks like I failed to update the readme to reflect the new model for specifying parameters -- sorry about that! Maybe change the "If necessary [...]" sentence to something like "If necessary, update parameters that require a reference genome id, fasta or gtf file to include the `variable` slot, containing one of `ASSEMBLY_ID`, `ASSEMBLY_FASTA_URL`, or `GENE_MODEL_URL`."
- I think it would be safer to create the updated workflow data by merging new values from the current workflows into the existing workflows, rather than merging existing values into the current workflows, as it is now. That way it's easier to avoid accidentally discarding new manually-written fields -- for example, since this PR was opened, we've added an optional `taxonomy_id` field, which would be overwritten by the script as it is currently.

mvdbeek:
I can update the readme. We can limit the update to fields that come from the upstream manifest, but we do have to overwrite the existing values to get new parameters, descriptions and names.

hunterckx:
> I can update the readme. We can limit the update to fields that come from the upstream manifest, but we do have to overwrite the existing values to get new parameters, descriptions and names.

@mvdbeek Thanks! Yeah, I was thinking that for existing workflows we should update only the necessary values, and make sure anything that isn't populated from the IWC manifest is left in place.

catalog/output/workflows.json
```diff
@@ -19,32 +29,72 @@
...
       },
       {
+        "parameters": [
+          {
+            "key": "Reference genome",
+            "variable": "ASSEMBLY_ID"
+          },
+          {
+            "key": "GTF file of annotation",
+            "variable": "GENE_MODEL_URL"
```
mvdbeek, line 90: @NoopDog This should be the missing parameter we talked about on the call today.

hunterckx (approved):
Looks good, thanks @mvdbeek!

PR #365, opened 2025-03-08
fix: use correct variant calling landing body even when gene model url is absent (#347)

### LINKED ISSUE #347: Populate assembly URL for assemblies without gene model URLs when launching Galaxy workflow
For assemblies without gene model URLs, populate assembly URL when launching the Galaxy workflow.

hunterckx:
Closes #347


mvdbeek (approved):
Thanks!

PR #368, opened 2025-03-09
fix: avoid assembly duplication due to duplicate taxonomy ids and add checks for duplicate entity ids (#367)

### LINKED ISSUE #367: Assemblies are duplicated
Due to an oversight in implementing batched API requests for getting taxon info, multiple rows may be returned for the same taxon when that taxon's ID appears multiple times in the input, which leads to duplication of assemblies when the assembly info is joined with the taxon info

hunterckx:
Closes #367


NoopDog (approved):

NoopDog:
Tanks @hunterckx

PR #369, opened 2025-03-10
feat: allow limiting workflows by taxonomy id (#356)

### LINKED ISSUE #356: Update workflow schema to support matching on organism
Add approch from parent ticket #355 

From #355: 

"As for how to map them in the schema, I'd like to be able to pick an arbitrary ncbi tax Id and let its children use the workflow too. I'm not convinced we need the root node toggle thing, provided everything won't break if I pass a tax id for a genus or some other non-organism thing. (Note: the arbitrary tax ids in this context seem an important thing to qc via an automated script and report)"

#### Approach

1. Add a field to the schema "tax_id" with a description and adjust he validations accordingly. The id can be at any level. e.g. kingdom, phylum, class, order.." 
2. On the organisms page, we will check that any organism-specific restriction on a workflow matches the given organism or any of its ancestors. 
3. Update the workflow matching on the assembly detail page to match a specific organism and descendants.

hunterckx:
Closes #356

A workflow can be limited to a certain taxon by specifying a taxonomy ID for the `taxonomy_id` field, e.g.:
```yaml
- trs_id: "foo"
  categories:
    - "TRANSCRIPTOMICS"
  workflow_name: "bar"
  workflow_description: "baz"
  ploidy: "ANY"
  taxonomy_id: 123
```


app/components/Entity/components/AnalysisMethodsCatalog/analysisMethodsCatalog.tsx
```diff
@@ -3,26 +3,23 @@ import workflows from "../../../../../catalog/output/workflows.json";
...
+  BRCDataCatalogGenome,
+  Workflow,
+} from "app/apis/catalog/brc-analytics-catalog/common/entities";
 
-export const AnalysisMethodsCatalog = ({
-  assemblyPloidies,
-  geneModelUrl,
-  genomeVersionAssemblyId,
-}: Props): JSX.Element => {
+export const AnalysisMethodsCatalog = ({ assembly }: Props): JSX.Element => {
```
hunterckx, line 11: This component now uses four properties from an assembly, so I've updated it to just take the assembly itself as a prop rather than taking the four properties separately

catalog/schema/generated/organisms.json
```diff
@@ -5,7 +5,7 @@
             "description": "Info for an organism.",
             "properties": {
                 "ploidy": {
-                    "description": "The organism's ploidy.",
+                    "description": "The ploidies that the organism may have.",
```
hunterckx, line 8: Apparently the files generated from the schema haven't been updated since this description was changed, so the change is showing up now

catalog/schema/scripts/gen-typescript.py
```diff
@@ -1,8 +1,10 @@
 import sys
 from linkml.generators.typescriptgen import TypescriptGenerator
 
+# The source for the base TypeScript generator can be found at https://github.com/linkml/linkml/blob/main/linkml/generators/typescriptgen.py
+
 """
-LinkML TypeScript generator extended to handle enums more correctly.
+LinkML TypeScript generator extended to handle enums and optional fields more correctly.
```
hunterckx, line 7: Apparently `null` is a valid value for optional fields, which was not reflected in the generated TypeScript, so I've updated the generator

hunterckx:
Thanks for the feedback @d-callan! I'll go ahead and merge this, but I've assigned myself #375 to update the descriptions to something more like your example

PR #372, opened 2025-03-10
chore: Remove issue number check from husky (#363)

### LINKED ISSUE #363: Can we please disable the extremely opinionated commit message checker
I don't want to be forced to reference an issue and remember numbers on every commit. I am aware of the `--no-verify` flag but then actually useful linting is being skipped.

dannon:
This closes #363.  It drops the issue number check, but leaves the commitlint validation.  If we want to follow up with a github workflow enforcing that PRs reference an issue, that's fine.

Also adds .venv to prettierignore.

### DIFF
--- a/.husky/commit-msg
+++ b/.husky/commit-msg
@@ -2,16 +2,6 @@
 
 echo "Checking commit message"
 
-# Check for issue number
-
-commit_regex='#[0-9]'
-error_msg="Aborting commit. Your commit message is missing a '#xxx' reference to a corresponding GitHub issue. Add '--no-verify' to your git commit command if you wish to skip this check."
-
-if ! grep -iqE "$commit_regex" "$1"; then
-    echo "$error_msg" >&2
-    exit 1
-fi
-
 # Check for Conventional Commits format
 
 npx commitlint --config ./commitlint.config.js --edit $1


dannon:
I could, but I don't think we should?

Each commit of a PR may be a different type, whether fix/enhancement, etc.,  so we should defer to individual PRs for that.  That's what someone's going to be looking at in git blame to see context, not a PR title.

And, for issue number, the body text of the PR is where we should reference an issue w/ a `closes <blah>` I think?  That's where github expects to find it (https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword), either in the PR description or a commit message itself.

NoopDog:
Hi @dannon the PR titles end up on the release notes. The commits are squashed when the PR is merged.  So... the PR title needs to indicate the type of change and what issue the PR is for. This would be the check that goes into the GitHub action. 

We are looking to be able to have releases with notes like:

image

NoopDog:
Oh and the PR title is used as the default title  for the merge commit. Thats why the PR title matters.

dannon:
Ahh, we squash.  I think that's unfortunate, but unless we're changing that too it makes sense to just decorate the PR title, that's fine.

NoopDog:
Awesome, thanks.

dannon:
Considering the release notes, though -- I do think they would also look cleaner and less confusing they were simply:

- add back all assemblies (#323)
- add filters for taxonomic ... (#301)

with only the PRs listed, not the issues and merge commits.  They simply aren't necessary and the PR contains all this information in the form anyone already on github reading the notes is going to want to consume.

hunterckx (approved):

PR #378, opened 2025-03-11
chore: replace HTML scraping with API calls (#346)

### LINKED ISSUE #346: Update gene model URL determination to use API instead of parsing HTML
#### Need
Remove HTML scraping from GTF workflow.

#### Approach
As per [this comment](https://github.com/galaxyproject/brc-analytics/issues/314#issuecomment-2695541156):

> use /list/files?genome=GCA_021951015.2 to find the GTF files at UCSC. For example, https://genome.ucsc.edu/list/files?genome=GCF_000002765.6.

mvdbeek:
Closes #346


catalog/build/intermediate/genomes-from-ncbi.tsv
```diff
@@ -1,773 +1,773 @@
...
-CBS 160.54	1279043	GCF_000365165.1	True	Scaffold		28991111	19.0	3197266.0	4.0	80.0x	54.5	Full annotation	GCA_000365165.2	Cladophialophora carrionii	86049	1,131567,2759,33154,4751,451864,4890,716545,147538,716546,147545,451870,34395,43219,82105,86049,1279043	Ascomycota	Eukaryota	Fungi	Ascomycota	Eurotiomycetes	Chaetothyriales	Herpotrichiellaceae	Cladophialophora	Cladophialophora carrionii	Cladophialophora carrionii CBS 160.54	https://genome.ucsc.edu/h/GCF_000365165.1	GCA_000365165.2	GCF_000365165.1	https://hgdownload.soe.ucsc.edu/hubs/GCF/000/365/165/GCF_000365165.1/genes/GCF_000365165.1_Clad_carr_CBS_160_54_V1.ncbiRefSeq.gtf.gz
-CBS 101466	1220924	GCF_000365145.1	True	Scaffold		28781303	19.0	2379673.0	5.0	132.0x	54.0	Full annotation	GCA_000365145.2	Cyphellophora europaea	293227	1,131567,2759,33154,4751,451864,4890,716545,147538,716546,147545,451870,34395,1341112,226991,293227,122
```
mvdbeek, line 2: The GTF urls didn't change. I don't know what's up with `Eukaryota` missing from the diff.

NoopDog: Thanks, will take a look!

hunterckx (changes requested):
Thanks @mvdbeek! A few small changes before we merge, please:
- On main, I've addressed the issue with disappearing superkingdoms -- if you rebase and re-run the script, those missing values should be fixed
- The version number in `catalog/build/py/package/setup.py` should be updated (I think this would qualify as a patch according to Semantic Versioning so just an update to `1.5.1`?)
- Because getting the gene model URLs can take some time, I think it would be good to give some feedback when it's happening -- not necessarily printing a line for every URL like before, but it might be handy to have something like a `print("Fetching gene model URLs")` at the beginning of the `add_gene_model_url` function

hunterckx (approved):
Looks good!

PR #387, opened 2025-03-13
feat: adding niaid biodefense viruses (#383)

### LINKED ISSUE #383: adding niaid biodefense pathogens: viruses
see parent issue for details

d-callan:
resolves #383 , though depends on the assemblies listed in the qc report as not in ucsc first finding their way there. im unsure the relative priority of that currently. ill leave as draft until that becomes more clear.

probably the list of assemblies could be improved or refined also. i suspect ive got more norovirus and friends than needed for ex. i can look at that (or anything else) again if wed like.

its also very likely there are ppl here w more educated opinions than myself about some of these organisms and things like which assemblies make sense for them etc. id be very happy for feedback in that regard.


d-callan:
we now have all of these in the brc hub at ucsc, thanks to hiram!

two things happened along the way:
1. i noticed some of these assemblies i tried to add were creating duplicate entries in the intermediate tsv file. i think it was once for genbank and once for refseq iirc. i think ive fixed that here.
2. i couldnt bear the gene model api thing any longer. it was taking 40+ mins and hiram had told me he had a file i could just pull down, so i did that instead. 

one note for the future. id like a stable url i can piont hiram at, that he can routinely/ automagically look to see if ive got new assemblies on the list for him. can we leave this branch after merge and reuse it, or what do we want to do otherwise to support this?

package.json
```diff
@@ -82,6 +82,7 @@
     "next-mdx-remote": "^4.2.0",
     "prettier": "^3.4.2",
     "prettier-plugin-organize-imports": "^4.1.0",
+    "ruff": "^1.5.4",
```
hunterckx, line 85: Looks like the NPM package `ruff` has been mistakenly installed? The changes to `package.json` and `package-lock.json` can likely be reverted

hunterckx (changes requested):

hunterckx (approved):
Cool, thanks @d-callan!

natefoo:
Indexes are built and available on usegalaxy.org and http://datacache.galaxyproject.org/brc/data/genomes/, with the exception of `GCA_001021295.1_ViralMultiSegProj286006`, which has a RefSeq ID (GCF_001021295.1) but is keyed at UCSC under the GenBank ID (GCA_001021295.1).

Issue #357, opened 2025-03-05
Add organism specific workflows


NoopDog:
Add workflows determined from parent ticket #355, after the schema supports thier addition.

PR #412, opened 2025-03-24
feat: adding organism specific workflows (#357)

d-callan:
a wip, but i have some questions so thought id make a pr to talk around. 

right now i added two workflows for covid (there are others, some for covid, 1 or 2 for mpox or flu.. but they all were suffering various issues, so i started w what was working). the questions i have are:

1. do we want to add the variant calling category in iwc for these? right now they are assigned 'SARS-CoV-2' in iwc which i map to 'Variant Calling' in brc, but that seems bad. it wasnt obvious to me how to update iwc for this..
2. looking at some of these in galaxy, i can see places where we tell users they can fetch input files from a public history of wolfgang's.. since we have made an effort before to pre-populate input params for ppl, do we want to try to do the same here? and if thats easy enough to do, how do i do it?


d-callan:
believe were adding the variant calling category to these workflows in iwc

catalog/source/workflows.yml
```diff
@@ -685,6 +685,194 @@ workflows:
...
+          class: Collection
+          ext:
+            - fastqsanger
+            - fastqsanger.gz
+      - key: NC_045512.2 FASTA sequence of SARS-CoV-2
+        type_guide:
+          class: File
+          ext:
+            - fasta
+            - fasta.gz
```
mvdbeek, line 840:
```suggestion
      - key: NC_045512.2 FASTA sequence of SARS-CoV-2
        variable: ASSEMBLY_FASTA_URL
```
is probably what you would want to do ?

catalog/source/workflows.yml
```diff
@@ -685,6 +685,194 @@ workflows:
...
+    taxonomy_id: 694009
+    parameters:
+      - key: Paired Collection
+        type_guide:
+          class: Collection
+          ext:
+            - fastqsanger
+            - fastqsanger.gz
+      - key: NC_045512.2 FASTA sequence of SARS-CoV-2
+        variable: ASSEMBLY_ID
```
mvdbeek, line 818:
```suggestion
        variable: ASSEMBLY_FASTA_URL
```

catalog/source/workflows.yml
d-callan: Yes, I'm getting there..

catalog/source/workflows.yml
```diff
@@ -685,6 +685,194 @@ workflows:
...
+      - key: ARTIC primer BED
```
mvdbeek, line 780: I think there are hardcoded urls to this bed file ? @hunterckx it would be good if we can directly specify parameters here.
Ideally we'd write:
```suggestion
      - key: ARTIC primer BED
         value:
           ext: bed
           src: url
           url: https://zenodo.org/record/4555735/files/ARTIC_nCoV-2019_v3.bed
```

catalog/source/workflows.yml
```diff
@@ -685,6 +685,194 @@ workflows:
...
+        type_guide:
+          class: File
+          ext: bed
+      - key: ARTIC primers to amplicon assignments
+        type_guide:
+          class: File
+          ext: tabular
```
mvdbeek, line 787: Same here:
```suggestion
        value:
          ext: tabular
          src: url
          url: https://zenodo.org/record/4555735/files/ARTIC_amplicon_info_v3.tsv
```

catalog/source/workflows.yml
```diff
@@ -685,6 +685,194 @@ workflows:
...
+          class: integer
+      - key: Maximum read length
+        type_guide:
+          class: integer
+      - key: NC_045512.2 FASTA sequence of SARS-CoV-2
+        type_guide:
+          class: File
+          ext:
+            - fasta
+            - fasta.gz
```
mvdbeek, line 752:
```suggestion
        variable: ASSEMBLY_FASTA_URL
```

d-callan:
> If you have a dockstore username I can add you so you can do that, but also we want to turn this around and make the IWC the source of truth.

@mvdbeek my dockstore username should be d-callan :pray:

d-callan:
sry @mvdbeek one more q for you.. do we think it makes sense to have [this](https://dockstore.org/workflows/github.com/iwc-workflows/sars-cov-2-pe-illumina-artic-ivar-analysis/SARS-COV-2-ILLUMINA-AMPLICON-IVAR-PANGOLIN-NEXTCLADE:main?tab=info) in the variant calling collection as well?

d-callan:
and [this one](https://iwc.galaxyproject.org/workflow/%23workflow%2Fgithub.com%2Fiwc-workflows%2Fpox-virus-amplicon%2Fmain/) maybe shouldnt be? im starting to feel we need a new workflow category like 'consensus sequences' or similar. and some of these new ones belong in that one either instead of or in addition to the variant calling one.

mvdbeek:
I've edited the collections on dockstore, if you run npm run iwc-manifest-to-workflows-yaml you should get the updated workflows.

> tools having changed since he was last saved.

difficult getting that old tool installed (https://github.com/galaxyproject/usegalaxy-tools/pull/944), i'll take a look tomorrow

d-callan:
so this ended up going places.. its doing enough interesting things now id like a review and merge w what workflows are here. we can always add more later.

things this ended up doing:
1. adding a workflow for mpox, and three for covid
2. adding a workflow category for consensus sequences
3. adding support for explicit url specs in workflows yml

d-callan:
(i dont love the explicit mapping of iwc virology collection to brc consensus sequences.. but so far everything in that iwc collection actually does look to have at least something to do w consensus sequences and i think requiring a mapping between the iwc collections and brc categories in the py script in order to get new workflows in the source yml is worth thinking more about separately anyhow.)

NoopDog:
Thanks so much for this, @d-callan and @mvdbeek. Reviewing!

hunterckx (approved):
Looks good to me! I've taken the liberty of simplifying the types slightly, removing the unnecessarily broad type from `WorkflowLandingsBodyRequestState` so that URL workflow parameters don't need assertions in order to be properly type-checked (I will also note: something I've recently learned of is TypeScript's `satisfies` keyword, which might be preferred over `as` if we need such a thing in the future)

PR #413, opened 2025-03-24
fix: include ncbiGene.gtf.gz file if there's no refSeq gtf (#414)

mvdbeek:


catalog/output/ncbi-taxa-tree.json
mvdbeek, line 1: I'm gonna back this out, I guess it's just the sort order that's not stable ?

mvdbeek:
@NoopDog @hunterckx can we get this deployed so that @nekrut can check that this works correctly ?

d-callan:
were the output json files actually updated?

mvdbeek:
good point, i only ran build-files-from-ncbi

PR #441, opened 2025-03-31
feat: add check for ploidy to the assemblies qc report (#366)

### LINKED ISSUE #366: On assemblies QC report check for assemblies without ploidy
This would find assemblies where the organism is not in the organism yaml or the organism yaml did not have ploidy.

d-callan:
resolves #366 

i did test it a bit by removing an entry from organisms yml and running locally, before turning it back for the pr.


hunterckx (approved):
Thanks @d-callan! I've bumped the version in `setup.py` and will merge in a moment

PR #443, opened 2025-04-01
feat: add launch galaxy button to workflow stepper (#437)

### LINKED ISSUE #437: Add Launch Galaxy button to workflow stepper
- Add the Accession value to `Reference Assembly` step.
- Introduce state for the side column.
- Add Launch Galaxy button to GTF Files step.

frano-m:
Closes #437.

image


frano-m:
> Using '!!geneModelUrls' only checks for undefined and may be truthy even if the array is empty. Consider using 'geneModelUrls && geneModelUrls.length > 0' to ensure that the step is active only when there are gene model URLs available.

I'm keeping this check here, we need to render the Step if `geneModelUrls.length > 0`. The check for `!!geneModelUrls` is intentional; the fetch is pending and we are still waiting for values. While this is occurring, we do not want to "open" the step up.

NoopDog (approved):
Nice @frano-m !

PR #447, opened 2025-04-01
feat: add loading ui to gtf step and stepper when launching galaxy (#446)

### LINKED ISSUE #446: Add loading UI to GTF step and stepper when launching galaxy
- Add loading UI to GTF step, while gene models are being fetched
- Add loading UI to stepper when launching galaxy

frano-m:
Closes #446.


frano-m:
> The loading state is determined by the truthiness of geneModelUrls. If geneModelUrls can be an empty array (which is truthy), this condition might not accurately reflect the intended loading state. Consider explicitly checking for 'undefined' or using a length check if it is an array.

We do not want to show the loading component, when `geneModelUrls` is an empty array. However, I have updated the code to check for `undefined` anyway!

NoopDog (approved):
Thanks @frano-m

PR #448, opened 2025-04-01
feat: add Release Drafter configuration and publish release workflow

dannon:
Per discussion at the weekly meeting, this is an option for release handling and notes generation.  Draft for discussion.


dannon:
(per discussion in group meeting, we'll try this and see how it goes)

Issue #459, opened 2025-04-04
Add SANGAR_READ_RUN variable handling to the haploid variant calling workflow and the Galaxy handoff


NoopDog:

MillenniumFalconMechanic:
Completed.

PR #460, opened 2025-04-04
feat: add stub of SANGER_READ_RUN workflow parameter variable (#459)

hunterckx:
Very minimal (just adding a new value to the enum and a placeholder case to the switch statement), but I think this is all that's useful to do unless we have a specific plan for what the relevant input and output to `paramVariableToRequestValue` needs to look like

Hopefully I spelled "Sanger" correctly -- I think it's misspelled in the ticket

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -58,7 +58,8 @@
     "jsdoc/check-alignment": "error",
     "jsdoc/check-param-names": "error",
     "react-hooks/exhaustive-deps": "error",
-    "sonarjs/redundant-type-aliases": "warn"
+    "sonarjs/redundant-type-aliases": "warn",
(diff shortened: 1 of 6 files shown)


.eslintrc.json
```diff
@@ -58,7 +58,8 @@
     "jsdoc/check-alignment": "error",
     "jsdoc/check-param-names": "error",
     "react-hooks/exhaustive-deps": "error",
-    "sonarjs/redundant-type-aliases": "warn"
+    "sonarjs/redundant-type-aliases": "warn",
+    "sonarjs/todo-tag": "warn"
```
hunterckx, line 62: Had to add this because I put in a TODO comment

NoopDog (approved):
Thanks @hunterckx

Issue #353, opened 2025-03-04
Review official list of outbreaks to track from NIAD


MillenniumFalconMechanic:

nekrut:
I will analyze the latest information from the CDC, NIAID, and WHO to prioritize the organisms in your list based on their importance as pathogens. This will be based on their impact on human health, emerging threats, and global disease burden. I'll update you once I have the findings.

# Ranked Priority Pathogens (Based on CDC, NIAID, and WHO Data)

1. **SARS-CoV-2 (COVID-19 coronavirus)** – *Highest Priority.* This novel coronavirus emerged in late 2019 and caused a global pandemic with unprecedented impact. It has infected over 760 million people and caused roughly 6.9 million reported deaths worldwide as of 2023 ([[Coronavirus disease (COVID-19) - World Health Organization (WHO)](https://www.who.int/news-room/fact-sheets/detail/coronavirus-disease-(covid-19)#:~:text=Coronavirus%20disease%20%28COVID,is%20thought%20to%20be%20higher)](https://www.who.int/news-room/fact-sheets/detail/coronavirus-disease-(covid-19)#:~:text=Coronavirus%20disease%20%28COVID,is%20thought%20to%20be%20higher)) – making it one of the deadliest outbreaks in modern history. Beyond the immediate health toll, COVID-19’s high transmissibility and potential for severe disease (especially in unvaccinated or high-risk groups) led to massive strain on healthcare systems and economies. Global authorities (WHO, CDC) continue to monitor and update strategies for this virus, keeping it at the top of priority lists for surveillance, vaccination, and research due to the ongoing threat of new variants and the sheer scale of its impact ([[Coronavirus disease (COVID-19) - World Health Organization (WHO)](https://www.who.int/news-room/fact-sheets/detail/coronavirus-disease-(covid-19)#:~:text=Coronavirus%20disease%20%28COVID,is%20thought%20to%20be%20higher)](https://www.who.int/news-room/fact-sheets/detail/coronavirus-disease-(covid-19)#:~:text=Coronavirus%20disease%20%28COVID,is%20thought%20to%20be%20higher)).

2. **Mycobacterium tuberculosis (Tuberculosis)** – *Critical Priority.* TB is a centuries-old pathogen that remains one of the leading causes of infectious death globally. In 2021 alone, an estimated 10.6 million people fell ill with TB and 1.6 million died from it ([
	Tuberculosis deaths and disease increase during the COVID-19 pandemic
[...]

PR #242, opened 2025-01-29
Update outbreaks.yml & include avian influenza

wm75:

PR #462, opened 2025-04-05
feat: add outbreak schema and initial outbreaks (#450)

### LINKED ISSUE #450: Add outbreak schema and populate with initial outbreaks info

d-callan:
resolves #450

supersedes #242, but lifts information from it

based on the list and info from #353 

makes some hopefully well-informed assumptions:
1. that for organism groups like the mucorales fungi, we want a single entry under outbreaks that can highlight specific descendants
2. we may want to stop showing certain entries without having to remove them, as the status of things change
3. we may want to display different types of resources about an outbreak slightly differently
4. we can figure out a way to meaningfully associate arbitrary taxonomic ranks (ex: 38574) with their descendant organisms and assemblies (outside of just the highlighted ones) if we would like
5. when the taxonomy id associated w an outbreak is actually mapped to an organism specifically (rather than a group of them) then we can highlight that organism in the same way we do the case of groups. that is to say, i dont need to duplicate the taxonomy_id under descendants to highlight

im very happy to revisit any of these as we start to actually use these and learn more about our needs. my aim was to put together something that seemed like it represented the most likely needs given the outbreaks list we already had in hand.

### DIFF
--- a/catalog/build/py/generated_schema/schema.py
+++ b/catalog/build/py/generated_schema/schema.py
@@ -92,7 +92,7 @@ class OutbreakPriority(str, Enum):
     HIGHEST = "HIGHEST"
     CRITICAL = "CRITICAL"
     HIGH = "HIGH"
-    MODERATE_HIGH = "MODERATE-HIGH"
+    MODERATE_HIGH = "MODERATE_HIGH"
     MODERATE = "MODERATE"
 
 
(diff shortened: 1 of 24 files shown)


NoopDog:
Thanks so much, @d-callan! We will review.

hunterckx (approved):
Nice -- thanks @d-callan!

PR #469, opened 2025-04-16
feat: return collection for `SANGER_READ_RUN` workflow parameter variable (#468)

### LINKED ISSUE #468: Return collection for `SANGER_READ_RUN` workflow parameter variable
See #466 for info on collection structure

May be best done based on `fran/456-paired-end-step`, which already has code for populating `SANGER_READ_RUN` parameters with a placeholder structure

hunterckx:
Closes #468

Based on `main` as opposed to (as suggested in the ticket) `fran/456-paired-end-step` because the latter has linting errors which would make things awkward

### DIFF
--- a/app/apis/catalog/brc-analytics-catalog/common/entities.ts
+++ b/app/apis/catalog/brc-analytics-catalog/common/entities.ts
@@ -1,14 +1,9 @@
+import { WorkflowUrlParameter } from "../../../../utils/galaxy-api/entities";
 import {
   ORGANISM_PLOIDY,
   WORKFLOW_PARAMETER_VARIABLE,
   WORKFLOW_PLOIDY,
 } from "./schema-entities";
-
-export interface WorkflowUrlParameter {
-  ext: string;
-  src: string;
-  url: string;
-}
 
 export type BRCCatalog = BRCDataCatalogGenome;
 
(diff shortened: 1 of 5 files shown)


hunterckx:
Thanks for the feedback @mvdbeek! I've updated it to use the run accession, and opened this for review -- could you take a look?

One thing I want to note is that I've set it up to use the file with `_1` in the name for the "forward" data, and the file with `_2` for the "reverse" data, which I don't actually know for sure is accurate. Any insight on that?

NoopDog:
Hi @mvdbeek can you take a look at @hunterckx's questions above? Thanks! D

mvdbeek:
> I've updated it to use the run accession, and opened this for review -- could you take a look?

Thanks, yes, that looks good.

> One thing I want to note is that I've set it up to use the file with _1 in the name for the "forward" data, and the file with _2 for the "reverse" data, which I don't actually know for sure is accurate. Any insight on that?

Not really, since the `;` is what the ENA API serves I would assume they provide the files in the correct order, so what I would do is just accept that order instead of guessing using the `_1` /  `_2` heuristic. We have some contacts to the ENA (@bgruening @wm75 do you know anyone we can contact about this question ?), the best option I guess is to clarify this.
The second best is to sample a large number of paired runs and see if the _1 / _2 heuristic holds up ?

mvdbeek (approved):

hunterckx:
Cool -- I'll go ahead and merge this and we can address forward/reverse determination in a separate ticket: #471

PR #476, opened 2025-04-28
feat: add simple feature flag defaulting viz to off, for now (#289)

dannon:
Per discussion, for now, this provides a simple feature flag defaulting homepage genome viz to off.

To play with it, demo it, whatever: `localStorage.setItem("show_assemblies_section", "true");`


dannon:
(or, if we're happy enough with it shipped as-is, we can totally just close this PR)

dannon:
I think we're just going to ship it, so I'll close this.

Issue #384, opened 2025-03-13
adding niaid biodefense pathogens: bacteria


d-callan:
see parent issue for details

PR #486, opened 2025-04-29
feat: adding niaid biodefense bacteria (#384)

d-callan:
leaving a draft until we have these in ucsc

some things that happened along the way:
1. i got annoyed by the order of taxa in the tree data for the landing page viz not being consistent and that should be fixed now w this
2. once we added more organisms, sometimes the request for taxonomy info was producing a 502. the spot this happened i broke into batches. 
3. i noticed the qc report for gene model urls was always saying it found everything, even when that was not possible. that should be fixed here as well


MillenniumFalconMechanic:
#384

hunterckx (approved):
Thanks!

Issue #494, opened 2025-04-30
Outbreaks  - Make title explicit in the schema, yamls and json


NoopDog:
#### Need

<img width="939" alt="Image" src="https://github.com/user-attachments/assets/dc628c1d-501b-42ef-a58f-1d6f85867e26" />

For the outbreak details page and cards, we need the title to be explicit. 
It may not be so straightforward to derive the titles from the taxonomy ID, as these can be groups. In any case, the manual title is more flexible.

#### Approach

- Add title to the outbreak schema
- Add titles to the outbreak yamls
- Update the processing to include the title in the output JSON as part of the data build!

d-callan:
i mean we could pull the titles directly from ncbi taxonomy, but then well get things like 'leish donovani species complex' when what were trying to highlight is probably 'visceral leish'. since we have to manually curate the outbreaks anyway, we may as well take the time to give them meaningful titles, is what i figure.

NoopDog:
Exact!  The intent of this ticket is to allow for manual specification of the titles.

PR #498, opened 2025-05-06
fix: outbreak titles should be explicit (#494)

d-callan:
we previously had names/ titles for outbreaks worked into the markdown description files. this should hopefully improve the names provided, and also make them available via a dedicated field.


hunterckx (approved):
Thanks @d-callan!

PR #501, opened 2025-05-08
feat: update data catalog to support filtering on outbreak taxa (#495)

### LINKED ISSUE #495: Add data to enable filtering by outbreak taxa
#### Requirements

In order to allow outbreaks to link to associated assemblies/organisms, we would like to:
- Ensure that the primary taxon for each outbreak appears (by name) in some field in the assembly and organism lists, so that it can be filtered for.
- Ensure that such a filter can be generated from an outbreak (i.e. we need to know the name to filter for and the field it appears in).

#### Approach

- Add a new field, with a name such as `otherTaxa`, to the build output for assemblies and organisms. This field will contain a list of names of the organism's taxa that fall outside of the taxonomic ranks we explicitly save (as determined by `taxonomic_levels_for_tree` in `build.py`). This should likely be limited to taxa referenced by outbreaks.
  - Add a parameter to `build_files` to specify the path to read `outbreaks.yml` from, in order to get the list of outbreak taxonomy IDs.
  - Update `build.py` to get names of outbreak-associated unranked taxa from NCBI, and save them in the `otherTaxa` field. We already save the IDs of each assembly's full taxonomic lineage in the `lineageTaxonomyIds` field, which could potentially be used to match with outbreak taxonomy IDs, although as it's a list (and currently stored as a comma-separated string) it might be awkward to work with -- maybe there's an alternative strategy?
  - Update the files under `catalog/build/ts` as appropriate to add `otherTaxa` to assemblies and organisms.
- Add two new fields, e.g. `taxonNameField` and `taxonName`, to the build output for outbreaks. These will specify the key (e.g. `otherTaxa`, `taxonomicLevelSpecies`, etc) and value, respectively, to use when building a filter for the assemblies or organisms list.
  - Update `build-outbreaks.ts` to add these fields. This may involve taking the built assemblies as input in order to look up which field to filter by. (Or else having the Python script save some additional info under `catalog/build/intermediate`?)

d-callan:
resolves #495 

some things of note:
1. only the primary tax id is under the new 'otherTaxa' field for the genomes
2. for the descendant taxa to highlight, im currently assuming/ checking they are species ids. thats not necessarily a requirement we want going forward, but i think it meets the current needs and seemed the simplest thing to get us started. what it means if ive understood things right is that we can potentially produce links like https://brc-analytics.dev.clevercanary.com/data/organisms/498019 as things are.
3. i also had to update the tax id for covid, bc ncbi changed some stuff


catalog/build/py/package/catalog_build/build.py
```diff
@@ -968,6 +1100,53 @@ def build_files(
...
+        def get_other_taxa(lineage_ids):
+            taxa = []
+            for tax_id in lineage_ids:
+                if tax_id in outbreak_taxonomy_ids:
+                    # Check if this taxon's rank is already covered by taxonomic_levels_for_tree
+                    rank = outbreak_taxon_rank_map.get(str(tax_id), "").lower()
+                    if rank not in taxonomic_levels_for_tree:
+                        # Use the taxon name instead of the raw ID
+                        taxa.append(outbreak_taxon_name_map[str(tax_id)])
+            return taxa if taxa else None
```
hunterckx, line 1132: Presumably `taxa` should get converted to a comma-separated string here to match the expected format in `build-assemblies.ts` -- as it is, Pandas saves it as a Python literal

catalog/build/py/package/catalog_build/build.py
```diff
@@ -289,8 +313,8 @@ def get_species_tree(taxonomy_ids, taxonomic_levels, species_info=None):
                 ]
                 if "rank" in info["taxonomy"]:
                     taxon_rank_map[tax_id] = info["taxonomy"]["rank"]
-                else:
-                    print(f"rank not found for tax_id: {tax_id}")
+                # else:
+                #     # print(f"rank not found for tax_id: {tax_id}")
```
hunterckx, line 317: I assume this should be uncommented? (Or else removed entirely if there's something that warrants that)

catalog/build/py/package/catalog_build/build.py
```diff
@@ -351,8 +377,8 @@ def fetch_taxa_info(tax_ids, taxon_name_map, taxon_rank_map, description="taxa")
         taxon_name_map[tax_id] = report["taxonomy"]["current_scientific_name"]["name"]
         if "rank" in report["taxonomy"]:
             taxon_rank_map[tax_id] = report["taxonomy"]["rank"]
-        else:
-            print(f"rank not found for tax_id: {tax_id}")
+        # else:
+        #    print(f"rank not found for tax_id: {tax_id}")
```
hunterckx, line 381: (This commented section as well)

hunterckx (changes requested):
I see a few small adjustments to make as listed below -- otherwise, looks great! Thanks

d-callan:
so when i went to address the feedback, i had to rerun the py script and discovered that apparently now tax ids from ncbi are int rather than str.. so thats hopefully addressed here now also.

d-callan:
(note: im going to wait to run again the qc report for https://github.com/galaxyproject/brc-analytics/pull/486 until this is merged so i can get the fixes to address ncbi updates there also)

hunterckx (changes requested):

catalog/build/py/package/catalog_build/build.py
```diff
@@ -968,6 +1113,54 @@ def build_files(
         species_info, species_name_info, taxonomic_group_sets, taxonomic_levels_for_tree
     )
 
+    print(f"Found {len(outbreak_taxonomy_ids)} outbreak taxonomy IDs")
+    # Add otherTaxa field with outbreak-associated taxa names
+    if outbreak_taxonomy_ids:
+        # Convert lineageTaxonomyIds from comma-separated string to list of integers
+        species_df["lineageTaxonomyIdsList"] = species_df["lineageTaxonomyIds"].apply(
+            lambda x: [int(id) for id in x.split(",")]
```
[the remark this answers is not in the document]
d-callan: i actually started on that road initially, making them ints. and then realized id have to change quite a bit and this pr was large already so decided to get out as fast as i could for now. 

didnt have much choice to catch though, things were breaking :(

d-callan:
:crossed_fingers: third times the charm hopefully lol

hunterckx (approved):
Cool, thanks @d-callan!

PR #538, opened 2025-05-21
feat: include hashes in paired collections sent to galaxy (#470)

### LINKED ISSUE #470: Include checksums in landing requests
This would allow us to detect issues with uploads, and importantly let us deduplicate repeated uploads. The ENA is serving md5's under `fastq_md5`. hashes can be included as an array in the `File` (or `src: url`) element of an upload request, see https://json-schema.app/view/%23/%23%2F%24defs%2FFileRequestUri/%23%2F%24defs%2FDatasetHash?url=https%3A%2F%2Fgist.githubusercontent.com%2Fmvdbeek%2F156d6a3bef855a0e34704c51dee9e303%2Fraw%2F8a7b4c5e40d8df9e6d8576ca593c723a12709d4a%2FDataOrCollectionRequest.json.

Here's an example from Galaxy's tests:
```python
                "WorkflowInput1": {
                    "src": "url",
                    "url": f"base64://MSAyIDM=",
                    "ext": "txt",
                    "hashes": [{"hash_function": "MD5", "hash_value": "5ba48b6e5a7c4d4930fda256f411e55b"}],
                },
```

hunterckx:
Closes #470

Passing the hashes through from the stepper to the Galaxy API requires some changes to how the steps work:
- `ConfiguredInput` now contains a specific set of fields which may have different types, rather than containing arbitrary entries all of the same type.
- Labels are no longer stored in `ConfiguredInput` -- instead, the view model builder for the side panel maps over the `STEPS` array, and for each step uses the `label` field to determine the entry label, and the function in the `renderValue` field to determine how the value is displayed. (Note: I believe this is the one place where the new system is less powerful than the old one -- with the new system, there's a one-to-one correspondence between steps and side panel entries.)
- Due to the change in how labels are used, `onConfigure` just takes a key and a value. Since the values may be different types, this requires the type system to know what the key is, so keys are referenced via the step config constant rather than being passed to the step component as a prop.

### DIFF
--- a/app/components/Entity/components/ConfigureWorkflowInputs/components/Main/components/Stepper/stepper.tsx
+++ b/app/components/Entity/components/ConfigureWorkflowInputs/components/Main/components/Stepper/stepper.tsx
@@ -9,15 +9,14 @@ export const Stepper = ({ workflow, ...props }: Props): JSX.Element => {
   const { activeStep, onContinue, onEdit } = useStepper(steps);
   return (
     <StyledStepper activeStep={activeStep} {...STEPPER_PROPS}>
-      {steps.map(({ key, label, Step, ...stepProps }, i) => {
+      {steps.map(({ label, Step, ...stepProps }, i) => {
         const active = activeStep === i;
         const completed = activeStep > i;
         return (
           <Step
-            key={key}
+            key={i}
             active={active}
             completed={completed}
-            entryKey={key}
             entryLabel={label}
             index={i}
             onContinue={onContinue}
--- a/app/components/Entity/components/ConfigureWorkflowInputs/components/Main/components/Stepper/stepper.tsx
+++ b/app/components/Entity/components/ConfigureWorkflowInputs/components/Main/components/Stepper/stepper.tsx
@@ -9,7 +9,7 @@ export const Stepper = ({ workflow, ...props }: Props): JSX.Element => {
   const { activeStep, onContinue, onEdit } = useStepper(steps);
   return (
     <StyledStepper activeStep={activeStep} {...STEPPER_PROPS}>
-      {steps.map(({ label, Step, ...stepProps }, i) => {
+      {steps.map(({ description, disabled, label, Step }, i) => {
         const active = activeStep === i;
         const completed = activeStep > i;
         return (
@@ -17,12 +17,13 @@ export const Stepper = ({ workflow, ...props }: Props): JSX.Element => {
             key={i}
             active={active}
             completed={completed}
+            description={description}
+            disabled={disabled}
             entryLabel={label}
             index={i}
             onContinue={onContinue}
             onEdit={onEdit}
             workflow={workflow}
-            {...stepProps}
             {...props}
           />
         );
--- a/app/utils/galaxy-api/galaxy-api.ts
+++ b/app/utils/galaxy-api/galaxy-api.ts
@@ -3,6 +3,7 @@ import { WorkflowParameter } from "../../apis/catalog/brc-analytics-catalog/comm
 import ky from "ky";
 import { GALAXY_ENVIRONMENT } from "site-config/common/galaxy";
 import {
+  EnaPairedReads,
   WorkflowLanding,
   WorkflowLandingsBody,
   WorkflowLandingsBodyRequestState,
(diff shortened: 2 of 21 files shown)


hunterckx:
@mvdbeek, could you review this, in particular for the actual Galaxy-related changes in `galaxy-api.ts`?

@frano-m, let me know if the stepper changes look reasonable!

Thanks!

app/utils/galaxy-api/galaxy-api.ts
```diff
@@ -124,17 +130,25 @@ function paramVariableToRequestValue(
...
+  forward: EnaFileInfo;
+  reverse: EnaFileInfo;
   runAccession: string;
 } {
-  let forwardUrl: string | null = null;
-  let reverseUrl: string | null = null;
+  const splitUrls = enaUrls.split(";");
+  const splitMd5Hashes = enaMd5Hashes.split(";");
+  if (splitMd5Hashes.length !== splitUrls.length)
+    throw new Error("Hash list has different length than URL list");
```
mvdbeek, line 147: Is that gracefully handled somewhere in the UI ? Will a user have an alternative ? We could also just render a warning but allow using the files anyway, it's not a critical requirement for functionality, but a nice to have that allows us to do some optimizations on the galaxy backend.

app/utils/galaxy-api/galaxy-api.ts
```diff
@@ -124,17 +130,25 @@ function paramVariableToRequestValue(
...
  * Get run accession and full URLs for the given paired run URLs from ENA.
  * @param enaUrls - Concatenated paired run URLs, as provided by ENA.
- * @returns accession and URLs for forward and reverse runs.
+ * @param enaMd5Hashes - Concatenated MD5 hashes of the files referenced by the URLs, as provided by ENA.
+ * @returns accession, URLs, and hashes for forward and reverse runs.
  */
-function getPairedRunUrlsInfo(enaUrls: string): {
-  forwardUrl: string;
-  reverseUrl: string;
+function getPairedRunUrlsInfo(
```
mvdbeek, line 136: generally speaking it might be good to place ENA parsing in a separate file ? I guess I wouldn't look for this in galaxy-api.ts ?

mvdbeek (approved):
Looks great, thank you!

app/utils/galaxy-api/galaxy-api.ts
hunterckx: There's no special handling for the different errors that might be thrown, but the page content will be replaced with an error message if one occurs, so the user will at least know that something has happened and be able to reload the page to try something else

app/utils/galaxy-api/galaxy-api.ts
hunterckx: Yeah, I agree -- I didn't do that in this PR because we already had `getPairedRunUrlsInfo` in `galaxy-api.ts` and moving it seemed out-of-scope, so I've just opened #547 to cover separating out the ENA stuff

app/components/Entity/components/ConfigureWorkflowInputs/components/Main/components/Stepper/stepper.tsx
```diff
@@ -9,21 +9,21 @@ export const Stepper = ({ workflow, ...props }: Props): JSX.Element => {
...
   return (
     <StyledStepper activeStep={activeStep} {...STEPPER_PROPS}>
-      {steps.map(({ key, label, Step, ...stepProps }, i) => {
+      {steps.map(({ description, disabled, label, Step }, i) => {
         const active = activeStep === i;
         const completed = activeStep > i;
         return (
           <Step
-            key={key}
+            key={i}
```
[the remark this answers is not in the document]
hunterckx: I think `i` should work fine here since it's an index in a constant and indices are established as a way to identify the steps elsewhere in the code

app/views/WorkflowInputsView/hooks/UseConfigureInputs/types.ts
```diff
@@ -1,21 +1,18 @@
...
+export type OnConfigure = (...p: OnConfigureParams) => void;
 
-export interface ConfiguredInput {
-  [key: string]: ConfigurationEntry;
-}
+/**
+ * Possible tuples containing a key and its associated value type.
+ */
+export type OnConfigureParams = {
+  [K in keyof Required<ConfiguredInput>]: [K, ConfiguredInput[K]];
```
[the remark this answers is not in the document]
hunterckx: These are not in fact equivalent

frano-m (approved):
LGTM thank you @hunterckx 🌭

PR #549, opened 2025-05-23
feat: update sunburst colors (#482)

### LINKED ISSUE #482: Landing page sunburst colors
Right now a categorical color map is used w a limited number of colors, one of which is gray. The limited colors means they repeat when there are a lot of children. And the use of gray would probably be better reserved for places where you want to de-emphasize something or similar. 

Making large categorical color maps that are color deficient friendly is hard. I was considering a continuous circular color map we could pull arbitrary numbers of 'equidistant' colors from as needed.

d-callan:
resolves #482

adds some fun color-deficient friendly color maps (happy to put those somewhere else if theres a better place) that we can use in this or any other vizs we like. also updates the sunburst to use one of those color maps without repeating colors or using gray for one of the top tier children.


hunterckx:
I wonder if we should include licensing information or more explicit attribution for the colors? Common wisdom seems to be that color palettes can't be copyrighted, so quite possibly this doesn't apply if it's the same colors but different code, but Crameri's palettes appear to be distributed with an MIT license: https://zenodo.org/records/8409685

d-callan:
i dont think its strictly necessary.. what did you have in mind?

PR #550, opened 2025-05-23
feat: generate taxa tree from assemblies and display accurate assembly count in sunburst (#522)

### LINKED ISSUE #522: Fix sunburst counts

hunterckx:
Closes #522

Updates the taxa tree to be built from the assemblies dataframe rather than the NCBI tree API; this makes it straightforward for the tree data to accurately reflect the catalog data in general (notably, there was previously some disconnect between which taxa appear in the tree and which appear in the catalog), as well as making assembly counts in particular straightforward to associate with tree nodes. Additionally, dataframe grouping makes the tree structure simple to derive.


hunterckx:
Could you give this a review @d-callan? I did end up removing the NCBI call for building the tree; hopefully it's clear why I think this approach is more practical, but do let me know if there's some shortcoming I may have overlooked! Thanks

d-callan (commented):
so ill admit i like this better than i thought i would. one thing that does make me nervous.. unless i misunderstood something, it seems like weve tied ourselves to the order of taxonomic levels being biologically relevant? i have mixed feelings about this: on the one hand that makes the logic here simple enough i trust it to build a valid tree, but i am not sure i trust anyone who might need to modify these levels later to know the order is important. at absolute min id say we need clear docs for this in the place the levels are actually defined.

d-callan:
actually ive realized i have one further thought on this. now that the assembly counts are accurate, they are more confusing. on the top of the child taxa list to the right of the sunburst we show counts of assemblies, but within the list we show counts of leaves (which are close but not quite the same as the organisms count). probably we should remove the count of leaf nodes. it wont be reliably guaranteed to match any other number we show on the site i dont think, isnt showing a preview of how many things well get if we click some link, and generally doesnt seem like its adding a lot of value to me.

d-callan:
Thanks. I said number tiers to try to catch cases where the levels were passed out of order, figuring whole levels would be missing say. But I guess what I really meant was maybe to check the actual levels/ranks claimed in the nodes and make sure all the levels desired actually exist somewhere in the total tree across all lineages.

d-callan (approved):
Thanks 🙏

NoopDog:
Thank you all !

PR #552, opened 2025-05-27
feat: move schemas and iwc workflows script to python package (#540)

### LINKED ISSUE #540: Add iwc_manifest_to_workflows_yaml.py to build package
See if it makes sense to add the iwc script to the build package, and do it if it does.

hunterckx:
Closes #540

- Since the IWC script as written references the schema, I've gone ahead and moved the schema definitions to the package, which is something we would presumably want to do anyway at some point
  - I've also ported some schema-related Bash scripts to Python and put them in the package
- In order to give the package greater prominence, I've moved it to be directly under `catalog` as `py_package`


d-callan:
Is this ready for review?

d-callan (approved):
my only real (but not urgent) comment is that while i like moving the py package up to a more prominent place, i kind of feel like it now seems like it needs a more descriptive name. 

outside of that, i feel like sharing that i had a couple fun and possibly issue-worthy thoughts while looking through and thinking about this:
1. i think i heard 'build_files_from_ncbi.py' say it wanted the configuration type stuffs split out into a different (not-python) file and the py package to gain the ability to find or be passed said config file.
2. id love if the workflows had a flow more similar to the assemblies. right now when i need to change something about the workflows, i have to look through a file w a lot of irrelevant entries and try to find the right one. then i have to read through a bunch of values im not supposed to modify to find the ones i need to. id rather a small file w only what im meant to add manually.
3. it feels like build.py is doing a lot.

d-callan:
Idk I guess I don't find the word catalog very descriptive either.. though if I'm real honest, to me, the fact we want this for multiple projects suggests it should maybe be its own repo? At min, how would you name it if it were might be an interesting thought experiment?

NoopDog (approved):
Thanks @hunterckx

PR #566, opened 2025-05-30
fix: counts in sunburst side panel represent assemblies (#564)

### LINKED ISSUE #564: Resolve conflict between counts in sunburst side panel
As of #550, the side panel in the sunburst displays two kinds of counts:
- In the list of subcategories, a parenthesized count of leaf nodes for each subcategory.
- In the label at the top, a count of assemblies for the current taxon.

The discrepancy between these numbers may be jarring, and it may not be clear what the leaf node count represents.

Since the count of leaf nodes doesn't have much meaning outside of the sunburst, we may want to remove the count entirely or replace it with the assembly count.

Image

d-callan:
resolves #564


d-callan:
:pray: @dannon

PR #599, opened 2025-06-19
feat: add IWC linkage from workflows

### LINKED ISSUE #583: Link from workflow description to IWC
For people who want more information about what a workflow is doing/going to do, it would be good to have a link from the "Select a Workflow" page to it's page at IWC.

dannon:
## Description

image

WIP, tinkering with display still, that button isn't right -- I think instead of a button we might want just a regular link below, anyway?  This is one option.  Adds iwcID to the schema for workflows and links them up to their README page on the IWC.

## Related Issue

closes #583


NoopDog:
We will get a mock designed for this shortly @dannon! Thx!

d-callan (approved):

Issue #499, opened 2025-05-08
make our data less fluid


d-callan:
sorry for that issue title, and the description thats about to follow.. i dont quite have a specific problem or solution yet, just a sense of one looming and some actions we might take that could potentially help that id like to put in peoples minds. here goes:

it seems to me we are too tightly bound to ncbi right now, to the point we effectively have no control of whats going to show on the site. it occasionally gets in the way when im working on something in the data catalog, but so far mostly in (a growing number of) small ways only. ill give an example of a few of these little issues though, and hopefully itll be clear how these can collectively add up to a real problem, particularly as we add more features and organisms.

1. they recently changed the ncbi taxon id that the covid assemblies are associated w. this means for us that the names of organisms will change, and their aliases, links to outbreaks weve defined will break, and the organism might even disappear from the site if i dont add the new tax id to our organisms yaml, etc unless i update all of that. its not a lot of work but i have no choice but to do that *right* now. i cant put it off until later, bc the second i run the data catalog for any even unrelated reason itll automatically update this.
     - prev: https://api.ncbi.nlm.nih.gov/datasets/v2alpha/taxonomy/taxon/2697049/name_report
     - new: https://api.ncbi.nlm.nih.gov/datasets/v2alpha/taxonomy/taxon/3418604/name_report
2. the name of c. auris changed and while everything still works and is even technically correct it happened silently without us even noticing at first and made an important organism difficult to find. (tangentially related, i think we need to stop trying to use the youth of the project as an excuse for things and instead just own it and go 'well look into it')
3. sometimes a new version of an assembly gets pushed out by ncbi, and then our qc report starts saying it cant find the old assembly on ncbi. but i dont think we want to be forced to update on their time tables. we should update, dont get me wrong, but we should have control over when we do that without losing organisms and assemblies from the site. 

i think there are two things weve sort of already talked about that might help:
1. making the data catalog more modular, so we dont have to rerun everything constantly and can just rerun the bits we actually need. that will help somewhat sometimes, maybe push off the problem a bit. its not a real fix though, bc there can still be times when i need to rerun a piece that ends up updating things i didnt intend to focus on in that moment.
2. the db were talking about for the sra metadata could expand in scope to include everything. it probably should at some point, but i understand thats a big thing with a lot of implications and means a lot of work. at the very least, if we arent already i think we should plan the sra metadata stuffs w the understanding that itll likely eventually grow quite a lot beyond that use alone.

[...]

PR #642, opened 2025-07-10
feat: add serotype and isolate info for assemblies

### LINKED ISSUE #555: consider adding serotype for viruses
itd be good to search for serotypes specifically, and to see them in the sunburst. ex H5N1 for flu (Taxonomy ID: 102793)
by a happy accident the sunburst had them before #550, but will lose them once thats merged.

### LINKED ISSUE #591: add isolate column to assemblies page
motivation being that for organisms w a lot of assemblies, its very difficult to tell them apart. some ppl might show up w a specific one in mind (P01) or p vivax for ex and be unable to find it. anything we can do to help differentiate them, the better.

d-callan:
## Description

this should help ppl distinguish and filter assemblies when there are many for a species. i can confirm that for at least a handful of important cases (flu and p falciparum for ex) this is a meaningful improvement. 

along the way i noticed we were filtering the organisms page on taxonomic ranks below species, which seemed odd to me. if i filter by strain, serotype or isolate and then get all assemblies for a species regardless of that filter it seems confusing. so i removed filters below species from the organism page.

also, we ran into a fun case of #499 type issues here.. this time a lot of viral species changed associated taxonomy ids. when i updated those ids and ran the data catalog again, there was another round of new id changes i had to contend w. so i got to iterate on that process until i caught up w ncbi apparently actively changing ids while i was working.

## Related Issue

closes #555 
closes #591


hunterckx (approved):
Looks good to me!

PR #644, opened 2025-07-11
fix: make assemblies more stable

### LINKED ISSUE #592: better handling of assemblies 'not found in ncbi' in the data qc-report
## Assemblies not found on NCBI

- GCA_020800235.1
- GCF_000002765.5
- GCF_000002825.2
- GCA_000003225.2
- GCF_000499545.1
- GCF_018416015.1
- GCF_023375885.1
- GCF_000857325.2

these presumably have newer versions, and we should update. but also, we should be able to keep them on the site w the older versions until we do, no? i somehow didnt quite realize before that these arent on the site bc of this and one of them is 3d7 it looks like, so im really not liking this.

d-callan:
## Description

we have an issue currently where older versions of assemblies drop off the site bc they arent being returned by the ncbi api. this modifies how we make requests so that well actually get those older assemblies weve requested. it also adds a few sections to the qc report so that well know when we have old or suppressed assemblies that we need to think about potentially updating or removing, etc.

along the way i noticed one assembly that actually had a typo that meant we were requesting a version that didnt yet exist, so fixed that. i also had some trouble w too large post requests/ responses and tried to make our helper for hitting the ncbi api a bit more robust to that

## Related Issue

closes #592


hunterckx (approved):
Thanks @d-callan!

Issue #588, opened 2025-06-16
bug: current release isn't linked to tag


dannon:
![Image](https://github.com/user-attachments/assets/af1d5a7c-1586-4fcb-a380-d7bef7c82643)

Reported by @NoopDog 

This *may* be because the linked release commit was https://github.com/galaxyproject/brc-analytics/commit/a83717aba66c92e380673cdf352cd3a5290f869c, which was a [no ci] followup to the actual tagged release when I had to tinker with the workflow.  We can validate this with a minor point release (which we need to do anyway) to include the typo fixes.

dannon:
Swing and a miss.  The updated workflow still has one commit past the tag, so we're still seeing `Unversioned-305d130` in prod.  This is because the PR that bumps the version happens *after* the release tag.  I'll adjust.

PR #656, opened 2025-07-15
chore: release tooling tweaks and doc updates

dannon:
## Description

fix: update version detection to be slightly more flexible/robust
docs: clarify tag usage in release publishing workflow

## Related Issue

another pass at https://github.com/galaxyproject/brc-analytics/issues/588


dannon:
Since this is release-specific I'm going to just push it through.

Issue #580, opened 2025-06-09
workflow stepper always shows gtf file input selection


d-callan:
it seems like we wont need this for all workflows no? like i wasnt actually expecting it for atac-seq, but saw it when i clicked that workflow just now.

Issue #659, opened 2025-07-16
Split SANGER_READ_RUN into single and paired versions


mvdbeek:
Since those are different inputs we'll need that for some of the workflows

d-callan:
related to #568

mvdbeek:
Also done in https://github.com/galaxyproject/brc-analytics/pull/662

Issue #652, opened 2025-07-15
Drive stepper steps from workflow config


NoopDog:
#### Need
The stepper currently renders all three steps it knows about for every workflow. These are

1. Reference Assembly
2. GTF Files
3. Paired-End Sequencing Data

This shows 3 in places where it is not required/desired.

Instead, we need to look at the workflow.yml parameters section, and only display workflow steps that are configured for the stepper (e.g. have a variable with a value we care about.

mvdbeek:
So the basics of this are implemented in https://github.com/galaxyproject/brc-analytics/pull/662 but I would rephrase this to say that we should build up both the stepper config and the values we send out in the landing request based on the workflow parameters, and track them based on the workflow input parameter name. There may for instance be optional inputs, or multiple inputs with separate fastqs (e.g. where samples need to be provided for treatment and control)

Issue #568, opened 2025-06-02
``SANGER_READ_RUN`` only substitutes `list:paired` parameters


mvdbeek:
We do however have workflows that can consume lists of unpaired files (all of the single end variants) or just single files.

d-callan:
it seems related, so ill mention, that the header 'Paired-End Sequencing Data' also seems hard coded maybe? probably makes sense to fix simultaneously

PR #662, opened 2025-07-16
feat: configure active workflows with stepper inputs

mvdbeek:
builds on #660 and #661.

Some of these workflows don't consume fastqs or GTFs or both, so I also implemented
https://github.com/galaxyproject/brc-analytics/issues/652 and some of these need single end collections, so I implemented https://github.com/galaxyproject/brc-analytics/issues/659

closes https://github.com/galaxyproject/brc-analytics/issues/580, https://github.com/galaxyproject/brc-analytics/issues/659, https://github.com/galaxyproject/brc-analytics/issues/568


mvdbeek:
I suppose that can happen at any point since those workflows aren't active

d-callan:
holy guacamole batman! hold please.. i need to go grab my thinking cap and put on some much higher energy music for this :joy:

app/utils/galaxy-api/entities.ts
```diff
@@ -32,7 +33,20 @@ export interface WorkflowUrlParameter {
   url: string;
 }
 
-// Narrow type specific to the one kind of collection we use -- might be worth defining a more general collection type if we need other kinds of collections
+// Narrow type specific to the two kinds of collections we use -- might be worth defining a more general collection type if we need other kinds of collections
```
d-callan, line 36: the relationship between these two things could be clarified. meaning, a lot of places were using single vs paired language but not here, and technically the paired one is also a list no? it leaves the names of the interfaces alone slightly confusing, until you think about it from galaxy perspective instead of just the context of this pr. i understood on my second read through, but still, even if all we do is improve the comment above.. future danielle's brain would appreciate it

app/components/Entity/components/ConfigureWorkflowInputs/components/Main/components/Stepper/components/Step/SequencingStep/components/UploadMyData/uploadMyData.tsx
```diff
@@ -4,11 +4,11 @@ import { Typography } from "@mui/material";
 import { TYPOGRAPHY_PROPS } from "@databiosphere/findable-ui/lib/styles/common/mui/typography";
 import { PAPER_PROPS } from "./constants";
 import { useEffect } from "react";
-import { STEP } from "../../step";
+import { PAIRED_END_STEP } from "../../step";
```
d-callan, line 7: this has been a lot to digest so maybe i didnt read it right, but this needs to be a bit cleverer no? or it thinks its not configured for single end workflows?

d-callan (commented):
a huge improvement! seriously, thanks @mvdbeek :superhero: 

> Not sure that the single / paired split for the ENA config is the best way to go about it. We'll have to eventually pass in props for organism, read type (short read vs long read), single-paired etc. that should control the default api filters we're using, so maybe passing along some kind of config object is a better idea.

i agree, but think all we need to do about it right now is file an issue and reference it here.

app/views/WorkflowInputsView/hooks/UseConfigureInputs/types.ts
```diff
@@ -11,7 +11,8 @@ export type OnConfigureParams = {
 
 export interface ConfiguredInput {
   geneModelUrl?: string | null;
-  readRuns?: EnaPairedReads[] | null;
+  readRunsPaired?: EnaPairedReads[] | null;
+  readRunsSingle?: EnaPairedReads[] | null;
```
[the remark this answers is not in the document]
mvdbeek: I wanted to limit the refactoring here, but you're right the interface is only

```
export interface EnaPairedReads {
  md5Hashes: string;
  runAccession: string;
  urls: string;
}
```
which is generic enough for single and paired reads. I'll rename it to EnaSequencingReads

app/components/Entity/components/ConfigureWorkflowInputs/components/Main/components/Stepper/components/Step/LaunchStep/launchStep.tsx
frano-m, line 1: Hey @mvdbeek! I noticed that the [STEP constant](app/components/Entity/components/ConfigureWorkflowInputs/components/Main/components/Stepper/components/Step/LaunchStep/step.ts) and [LaunchStep](app/components/Entity/components/ConfigureWorkflowInputs/components/Main/components/Stepper/components/Step/LaunchStep/launchStep.tsx) might not be used elsewhere in the codebase. Just wanted to check if these are still needed, or if they might be leftover code debt?

frano-m (approved):
LGTM! There might be some leftover code debt, and we might take a look at the UI with the "launch" button now outside of the last step (loading state looks a bit borked), but otherwise nice work!

Issue #123, opened 2024-10-07
Change NCBI wording on "Roadmap" page


nekrut:
For the BRC Analytics "Our Plan" page, could you tweak the wording about Gene pages? How about:
“Work with NCBI on evolution of gene pages to contain information valuable for understanding the structure and function of genes from a wide range of eukaryotes including eukaryotic pathogens and vectors.”

We want to make it clear that our plan (at least for NCBI) is not to recapitulate the VEuPathDB pages, plus NCBI's goals are not specifically constrained to this set of organisms.

NoopDog:
Complete ! Thx

murphyte:
thanks!

PR #671, opened 2025-07-23
fix: don't run eslint on anything that's in .gitignore

mvdbeek:
## Description

running eslint takes some time on my setup, probably since it descends into node_modules and the venv.
with these changes it's a lot faster

## Related Issue

If this addresses an existing GitHub issue, link it here (e.g., "Closes #123"). If no existing issue exists, consider creating one first to discuss the changes.


dannon:
Thanks, I noticed this too but never got around to addressing it :)

PR #678, opened 2025-07-23
feat: add display label for xeno refgene

### LINKED ISSUE #677: add display label for xenoRefGene gtf

d-callan:
## Description

i went w 'UCSC Xeno RefGene'. if others think theres something better to put there, let me know (maybe @maximilianh )

## Related Issue

closes #677

### DIFF
--- a/app/components/Entity/components/ConfigureWorkflowInputs/components/Main/components/Stepper/components/Step/GTFStep/utils.ts
+++ b/app/components/Entity/components/ConfigureWorkflowInputs/components/Main/components/Stepper/components/Step/GTFStep/utils.ts
@@ -47,6 +47,7 @@ export const getGeneModelLabel = (url: string): string => {
     [/\.augustus\./, "Augustus"],
     [/\.ncbiRefSeq\./, "NCBI RefSeq"],
     [/\.ncbiGene\./, "NCBI Gene"],
+    [/\.xenoRefGene\./, "UCSC Xeno RefGene"],
   ];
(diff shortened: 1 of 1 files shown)


d-callan:
yea, we could still do that at some point if we want. the data catalog checks for these, and will report if we dont have at least augustus. there are[ four there](https://github.com/galaxyproject/brc-analytics/blob/main/catalog/output/qc-report.md) currently that its complaining about. my thinking initially was to merge this as good enough until we sort those four. BUT if i look at homo sapiens on the dev site (one of the four) it doesnt load the gtf section at all. so..

i still think we could merge this and leave open the ticket you reference, that doesnt seem to hurt anything and is better than not having the display name but having the file listed there which is where we are now. but we could equally just skip to removing them, since that doesnt seem to make anything worse either when i thought previously that it would... i dont have strong opinions currently, so ill let you decide whether youd like to click approve or make a new pr. either way, ill talk to hiram about those four.

mvdbeek:
Oups, that's a good catch:

```
{
    "downloadTime": "2025:07:24T13:20:59Z",
    "downloadTimeStamp": 1753363259,
    "error": "can not find genome='GCF_000001405.40' for endpoint '\/list\/files'",
    "statusCode": 400,
    "statusMessage": "Bad Request"
}
```

i suppose there we'd use something like hg38 or whatever the cool kids do these days.

mvdbeek (approved):

Issue #679, opened 2025-07-24
Human gene models missing


mvdbeek:
if i look at homo sapiens on the dev site (one of the four) it doesnt load the gtf section at all. so..

_Originally posted by @d-callan in https://github.com/galaxyproject/brc-analytics/issues/678#issuecomment-3113340448_

maximilianh:
I think @nullmodel has an asmAlias for GCF_000001405.40 to hg38.

> returns 36792 items, almost half a megabyte in size and takes about 2.6
seconds to fetch.
Yes. Genomics on humans is a thing. :-)

mvdbeek:
Can that alias be used with the `/list/files` endpoint ? Otherwise I've added a similar alias in https://github.com/galaxyproject/brc-analytics/pull/681

PR #680, opened 2025-07-24
fix: add error handling to stepper steps

mvdbeek:
You can test this by going to the human assembly and see the GTF error. xref https://github.com/galaxyproject/brc-analytics/issues/679

Screenshot 2025-07-24 at 16 38 36

While not super actionable at least we're not showing the loading indicator indefinitely.


d-callan:
could this be made a warning, and we let ppl go to galaxy without it?

PR #681, opened 2025-07-24
fix: fix GTF stepper for human accession (#679)

mvdbeek:
closes https://github.com/galaxyproject/brc-analytics/issues/679 ... for now. I could see human being a frequently used test case for anyone discovering the site, so important that we fix this in a timely fashion.

Screenshot 2025-07-24 at 18 57 50


d-callan:
there were three other organisms in the list of things we didnt find gtf files for. are those suffering a similar problem? id like to understand how we got here better before doing this..

d-callan:
GCF_000001405.40
GCF_018416015.1
GCF_002986195.1
GCF_002986165.1

frano-m (approved):
LGTM @mvdbeek!

PR #694, opened 2025-08-12
feat: minimal agent/ chat

d-callan:
## Description

been playing w google genkit recently, decided to have a bit of an experiment in brc w it. this is the result of that effort over the last 12 hrs or so. i dont really anticipate this being merged, but i thought it was worth others playing w and could serve as a point of discussion for how we might approach this for real. 

current status: the llm (configure an api key) has access to the brc data catalog, some decent number of ncbi datasets api endpoints, and technically galaxy mcp though ive not played w that yet at all.. so far ive managed to get it to tell me what data is in the site, tell me which assembly was reference for an organism, what workflows are available and what they do, then for 3d7 i asked it to tell me how many genes in the assembly were annotated w specific things, or to list orthologs for a specific gene. thats already really interesting and helpful it seems to me, and like i said, i didnt even ask it to try to do anything w galaxy mcp yet though i did start stubbing something in for that.

future directions:  i figure we could wire in any number of fun things, so it could query sra and pubmed and suggest data based on knowledge of the study runs are associated w, or maybe do blast so ppl can paste in sequence and ask questions like 'what the hell is this'. ppl should start dreaming up interesting api endpoints we can work in from public resources. 

there are probably a lot of technical bits and bobs to consider here, i just did what got something working quickly. but a major thing to note is that there are some architectural implications to the fact the llm runs server side that i think will be obvious to anyone takes the time to look at this admittedly monstrous commit. (sry, like i kinda already said, started as a game and turned very cool too quickly to not share)

## Related Issue

mwahahahahahahahaha

ps his name is BioBuddy, fight me
biobuddy


d-callan:
in my head the ideal would be a biobuddy repo in the galaxy github, that multiple projects (brc, vgp, galaxy itself maybe) could use. i guess biobuddy would need a new face then :disappointed:

maximilianh:
Is there a way to play with this?

Issue #569, opened 2025-06-02
Workflows are outdated


mvdbeek:
Incredibly tedious to need an issue for this but here we go ...

mvdbeek:
Recent update in https://github.com/galaxyproject/brc-analytics/pull/707

PR #707, opened 2025-08-15
feat: add annotation workflow category and workflows

### LINKED ISSUE #651: incorporate amr workflow
https://iwc.galaxyproject.org/workflow/amr_gene_detection-main/

specific to bacteria

ive heard it mentioned amr is a niaid priority a couple times now, so this seems a good idea

d-callan:
## Description

ive heard amr mentioned as a priority a few times, and it seemed easy enough to add since there was already an iwc workflow for it and we have a good number of bacteria now. then once i added a workflow category for annotation workflows to pull it in i saw there were a handful of others looked kind of interesting too.

@mvdbeek please sanity check me. a couple of these use tools ive not heard of before or know fairly little about and there were a few places the input param keys felt ambiguous to me. also, two of the four additional workflows give warnings when i try to open them in galaxy about needing to modify something before i can use it:

- bacterial genome annotation
- annotation w helixer

i wonder if there is an easy way to send potential results of these to ucsc automagically, since we were talking about user contributed annotations the other day. maybe thats issue worthy..

## Related Issue

closes #651


d-callan:
think merging this would mean we can close #569 since versions will get updated here also

catalog/source/workflows.yml
```diff
@@ -463,10 +529,10 @@ workflows:
       - key: gtf
         variable: GENE_MODEL_URL
       - key: fastq PE collection
+        variable: SANGER_READ_RUN_PAIRED
```
mvdbeek, line 532: This shift in position is something prettier would revert, I think. Try running it ?

mvdbeek (approved):
Looks good to me, I'd just run prettier over it to be sure of the order of keys

d-callan:
thanks @mvdbeek , any thoughts on the two workflows that dont want to let me run them without modification?

mvdbeek:
Missing reference data I suppose, would have to copy over from .eu. Do you want to flag them as inactive for now until I get some time to do this ?

PR #749, opened 2025-08-25
fix: fix sequencing step study accession link (#748)

### LINKED ISSUE #748: Fix sequencing step study accession link

frano-m:
Closes #748.

This pull request simplifies and corrects the construction of the `href` property for study accession links in the ENA Sequencing Data table. The main change ensures that the link to the NCBI BioProject is only created when a value is present, and the URL is constructed correctly.

Improvements to link handling:

* Updated the `buildStudyAccession` function in `viewBuilders.ts` to construct the `href` property using a template string only when a value exists, ensuring accurate and conditional link generation.


MillenniumFalconMechanic (approved):
`LGTM, thanks F`

PR #789, opened 2025-09-05
fix: no results on ena picker when pre filters do not match (#778)

### LINKED ISSUE #778: No results on ENA picker when pre filters do not match
When using the ENA Picker.

### What I did

https://brc-analytics.dev.clevercanary.com/data/assemblies/GCF_004000055_1/workflow-github-com-iwc-workflows-haploid-variant-calling-wgs-pe-main-versions-v0-1

1. Go to Organsims page
2. Select Arthrobotrys flagrans
3. Select the one assembly with GCF_004000055.1 
4. Select variant calling
5. Select configure inputs
6. Select NCBI Refseq
7. Select Browse 69 sequences.

### What happened
Image

### What I expected
1. pop-up to show all  69 sequences and not prefilter, as no sequences match.

frano-m:
Closes #778.

This pull request refactors the logic for building validated column filters in the `CollectionSelector` utility by introducing a new approach that ensures only filter combinations that match at least one row in the table are applied. The previous implementation, which filtered out invalid columns and values individually, has been replaced with a stricter validation that checks the overall combination of filter values for validity. Additionally, a new utility function for value conversion was added.

Key changes:

**Column Filter Validation Logic:**

- Replaced the previous `buildValidatedColumnFilters` implementation with a new version that only applies preselected filters if at least one row in the table matches all filters (AND across columns, OR within a column). If no such row exists, no filters are applied. This ensures only valid filter combinations are used.

**Utility Functions:**

- Added a new `toStringArray` utility function to consistently convert values to string arrays, improving type safety and code clarity.

### DIFF
--- a/app/components/Entity/components/ConfigureWorkflowInputs/components/Main/components/Stepper/components/Step/SequencingStep/components/ENASequencingData/components/CollectionSelector/utils.ts
+++ b/app/components/Entity/components/ConfigureWorkflowInputs/components/Main/components/Stepper/components/Step/SequencingStep/components/ENASequencingData/components/CollectionSelector/utils.ts
@@ -4,6 +4,66 @@ import { ReadRun } from "../../types";
 import { PRESELECTED_COLUMN_FILTERS } from "./constants";
 
 /**
+ * Builds column filters from preselected values only if they are valid for the current data.
+ * A set of preselected filters is considered valid if at least one row in the table satisfies all column filters (AND across columns),
+ * where each column filter matches if the row contains at least one of the desired values for that column (OR within a column).
+ * If no rows satisfy the filters, an empty list is returned.
+ *
+ * Note: This does not mutate or filter out individual values; it validates the combination overall.
+ *
+ * @param table - Table instance.
+ * @param preselectedFilters - Record of column IDs to arrays of desired filter values.
+ * @returns Column filters state.
+ */
+function buildValidatedColumnFilters(
+  table: Table<ReadRun>,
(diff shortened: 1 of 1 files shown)


app/components/Entity/components/ConfigureWorkflowInputs/components/Main/components/Stepper/components/Step/SequencingStep/components/ENASequencingData/components/CollectionSelector/utils.ts
```diff
@@ -35,36 +96,12 @@ export function preSelectColumnFilters(
...
-    }
-
-    if (validValues.length === 0) continue;
-
-    columnFiltersState.push({ id: columnId, value: validValues });
-  }
-
-  return columnFiltersState;
+function toStringArray(value: unknown): string[] {
+  if (value == null) return [];
```
MillenniumFalconMechanic, line 104: Checking `==` is intentional here.

frano-m: LOL, no. I let Windsurf write that function 😱 .

frano-m, line 104:
```suggestion
  if (value === null) return [];
```

Issue #767, opened 2025-08-26
[GA2] Define minimum features available for the VGP meeting


Smeds:
We need to write down the features that must be available for the meeting at the end of September.

Must:
- [x] New name #754
- [x] Logo #756
- [ ] Show all VGP phase 1 assemblies ([github](https://github.com/VGP/vgp-phase1/blob/main/URL.download.table.tsv))
- [ ] Decide available workflows (we must also be able to use the stepper to start these) #755 
- [ ] Production site up and running (or a public available dev site) #759  
- [ ] Populated roadmap #791 (sub-issue of)

Nice to have:
- [ ] Populate Landing page
- [ ] Populate About page

MillenniumFalconMechanic:
Hi @Smeds, is there a ticket for "Show all VGP phase 1 assemblies"?

PR #795, opened 2025-09-08
feat: add text and components to GA2 landing page (#767)

Smeds:
## Description

First version of a landing page with the different sections and text that will be included

## Related Issue

Solves part of (#767)

## Design/organization question

@NoopDog @hunterckx Is this a good way of adding GA2 stuff? I have tried to follow how content/config was added in previous GA2 commits.

### Ex1
`./app/components/Home/components/Section/components/ga2/SectionAnalyticsAndData/sectionAnalyticsAndData.tsx` **vs** `./app/components/Home/components/Section/components/SectionAnalytics/sectionAnalytics.tsx`

where we basically only have difference at the following files:
1. `SectionAnalytics/sectionAnalytics.tsx` vs `ga2/SectionAnalyticsAndData/sectionAnalyticsAndData.tsx` text content updated
2.  `SectionAnalytics/components/AnalyticsTools/common/constants.ts` vs  `ga2/SectionAnalyticsAndData/components/AnalyticsToolsAndData/common/constants.ts` updated list `const ANALYTICS_TOOLS`

the other content/code of the `SectionAnalytics` and `SectionAnalyticsAndData` are the same.
### Ex2

`./app/components/Home/components/Section/components/ga2/SectionHelp` **vs** `./app/components/Home/components/Section/components/SectionHelp` here we could reuse the styles from `./app/components/Home/components/Section/components/SectionHelp/sectionHelp.styes.tsx` but I don't know if we want make GA2 stuff that dependent on BRC content?

@hunterckx @NoopDog  My question is whether I should update the GA2 code to reuse more of the BRC code and styles. The benefit would be less code duplication (always a plus), but the downside is that GA2 would become quite dependent on the BRC code.

image

image

image

### DIFF
--- a/app/components/Home/components/Section/components/ga2/SectionHelp/components/Questions/questions.tsx
+++ b/app/components/Home/components/Section/components/ga2/SectionHelp/components/Questions/questions.tsx
@@ -22,8 +22,8 @@ export const Questions = (): JSX.Element => {
           available, and downstream analysis results will be displayed. You can
           submit your own requests for additional datasets and features at{" "}
           <DXLink
-            label="help@brc-analytics.org"
-            url="mailto:help@brc-analytics.org"
+            label="help@genomeark2.org"
+            url="mailto:help@genomeark2.org"
           />
           .
         </MAccordionDetails>


frano-m (approved):
Looking fantastic! 🚀 

This is a LGTM, for now!
  
@NoopDog, @MillenniumFalconMechanic, and I are reviewing how to restructure to reduce duplication between GA2 and BRC. After the event, we’ll look to consolidate any duplicated components and improve sharing as part of the process.

Issue #753, opened 2025-08-26
[GA2] About page


Smeds:
The About page content of BRC Analytics looks like this and we will need to create a similar page for GA2

<img width="972" height="1570" alt="Image" src="https://github.com/user-attachments/assets/128b40f6-dacc-4d69-b2c4-e630edc84e1f" />

PR #821, opened 2025-09-10
feat: GA2 about page (#753)

Smeds:
## Description

First version of a about page with the different sections and text that will be included

## Related Issue

Solves part of (https://github.com/galaxyproject/brc-analytics/issues/753)

Anton is looking at the content! I will update the text according to his comments.


frano-m (approved):
LGTM thanks @Smeds! 🍰

Issue #841, opened 2025-09-17
Make organism selector more configurable


Smeds:
I’ve been experimenting with GenomeArk2 and added the organism selector to its landing page. While doing that, I identified several potential improvements/features I’d like to see. Most of these are focused on making certain parts more configurable:

1. logo part is hard coded to: `   .attr("href", "/logo/brc.svg")`
2. data loading is hard coded to `import data from "catalog/ga2/output/ncbi-taxa-tree.json";`

It would be great if we could specify which values to use in a config file, or alternatively use the appKey (e.g., appKey: APP_KEYS.GA2) to determine which value to select.

Also if we could make the **_starting level_** configurable somehow. For GenomeArk2 it looks like this:

<img width="472" height="315" alt="Image" src="https://github.com/user-attachments/assets/af34560b-7b18-484e-9e3e-7d252bfe259c" />

and I would like to be able to start at for example **Metazoa**, either that the selector zoom into that level or filter away the levels before this.

<img width="472" height="315" alt="Image" src="https://github.com/user-attachments/assets/392666e6-1b76-4753-9387-2297471c6191" />

It would be kinda nice to at least be able to easily configure the logo and data import for the VGP meeting in end of Sept.!

@dannon @NoopDog

PR #854, opened 2025-09-21
feat: make sunburst plot configurable

Smeds:
feat: make sunburst plot configurable by allowing developers to set center image, depth, and starting node via arguments

## Description

This change should make it possible to use the organism selector with GA2. For GA2, we need to change the center image and also need to be able to set a different starting node than the root.

## Related Issue

#841


Smeds:
@dannon here's the pull-request

Issue #868, opened 2025-09-26
[GA2] make MAX_READ_RUNS_FOR_BROWSE_ALL variable configurable


Smeds:
For the September 30th presentation, we’d like to use Zebra Finch (Taeniopygia guttata, tax ID: 59729) as an example. However, this organism has nearly 4,000 read runs. To handle this, we propose making the variable MAX_READ_RUNS_FOR_BROWSE_ALL [lhttps://github.com/galaxyproject/brc-analytics/blob/main/app/components/Entity/components/ConfigureWorkflowInputs/components/Main/components/Stepper/components/Step/SequencingStep/components/ENASequencingData/hooks/UseENADataByTaxonomyId/constants.ts ] configurable so we can set a custom value for GA2.

We’d like the option to show “Browse 3951 Sequences” even when there are more than 2,000 entries. Additionally, could performance be optimized? Displaying the table and applying filters is currently slow and makes working with this dataset difficult.

<img width="1438" height="1070" alt="Image" src="https://github.com/user-attachments/assets/577dd21c-0de8-4212-af3c-43f827f7db9e" />

PR #869, opened 2025-09-26
feat: change MAX_READ_RUNS_FOR_BROWSE_ALL from hard coded to configurable

Smeds:
## Description

For GA2 we would like to have a value for MAX_READ_RUNS_FOR_BROWSE_ALL that's different from the one used by BRC Analytics.

## Related Issue

#868


hunterckx (approved):
Looks good to me!

PR #892, opened 2025-10-06
feat: add backend infrastructure foundation

dannon:
## Summary
- FastAPI backend with Docker and nginx reverse proxy
- Redis caching infrastructure with async operations
- Health check and cache management endpoints
- Version syncing from package.json via Docker build args
- Frontend version display component with graceful fallback
- E2E tests for API health and version display

### DIFF
--- a/.github/workflows/run-checks.yml
+++ b/.github/workflows/run-checks.yml
@@ -2,6 +2,32 @@ name: Run checks
 on: [pull_request]
 
 jobs:
+  e2e-tests:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: "20.10.0"
+      - name: Cache npm cache
+        uses: actions/cache@v4
+        with:
+          path: ~/.npm
+          key: ${{ runner.os }}-node-${{ hashFiles('package-lock.json') }}
+      - name: Install dependencies
+        run: npm ci
+      - name: Install Playwright browsers
+        run: npx playwright install chromium
+      - name: Run e2e tests
+        run: npm run test:e2e
+      - name: Upload test results
+        uses: actions/upload-artifact@v4
+        if: failure()
+        with:
+          name: playwright-report
+          path: playwright-report/
+          retention-days: 7
+
   run-checks:
     runs-on: ubuntu-latest
     steps:


NoopDog (changes requested):
Hi @dannon, thanks so much for creating this. I would really love to get a minimal Python backend up and running. Looking at this, I see a few changes I would like to request. I will outline them here as they are a bit structural rather than line by line. So, in the order of the changed-file outline:

1.  Can you remove the `VersionDisplay` component? It looks like this duplicates or can be merged with the existing version display, and can be done in a separate PR.

2. In the `/backend` directory, can you move the contents there one level down in the folder hierarchy to make room for other backend applications, please? For example, I can see us implementing an "aI-service", a "api-service", and a "next-js-service" as discussed. Looks like this one is the `api-service`?

3. In` /scripts`, please co-locate docker-build.sh with the docker app that it will build.

4. In the top-level directory, can you
    1.  Move docker-compose.yml into the backend directory
    2. For nginx.conf, please move this to the backend directory and remove the section about hosting the Next.js static site content. CloudFront currently serves the static site via S3. No need to change this to get an API server running.
    3. ~~Remove Redis from this PR. Redis could always come in later if needed, but it doesn't seem to be required now, and it could be excluded from a minimal PR. Additionally, CloudFront caches items, depending on the cache control and TTL settings.~~ 



Thanks for adding playwright and updating the Jest config.

Thanks again for preparing this.

D

d-callan:
I think next js server needs a separate issue and pr, don't want to hold this up for that. More important, in my mind, to move on to talking about ncbi link out stuffs. 

As for dir tree, i see it as six of one half dozen of another. If we assume everyone is willing to discuss and iterate, it doesn't seem a blocker? 

Let's try to get this done please.

d-callan:
ok, i added an attached ticket for our future selves sake. when i say that i mean i pointed claude at the pr and told it to draft content for an accompanying ticket that summarizes where we landed and make it sounds like reqs lol then i just added the open questions section and a note to the dir structure section :joy: hopefully that makes it a little easier to parse if needed later. but also, maybe, can serve as a thing we can use to facilitate conversation about what our process should look like going forward. but we can talk about that later. so..

@dannon please look at the ticket and if claude did a bad job of anything feel free to update it to better reflect the current state of this pr, and then merge when youre ready. :tada: 

thanks again everyone :bouquet:

Issue #783, opened 2025-09-03
Link form NCBI to BRC Analytics


NoopDog:
### Summary  
NCBI and BRC Analytics are coordinating to improve cross-linking between their respective platforms.  

### Current State  
- **NCBI Assembly Pages** (e.g. [example](https://www.ncbi.nlm.nih.gov/datasets/genome/GCA_001008285.1/)) and **NCBI Organism Pages** (e.g. [example](https://www.ncbi.nlm.nih.gov/datasets/taxonomy/73230/)) currently provide assembly- and taxonomy-specific data from NCBI.  
- **BRC Analytics** already links out to NCBI from its assembly pages, allowing users to easily navigate to the canonical NCBI records.  

### Proposed Change  
- **NCBI** has offered to add reciprocal links **from its assembly detail pages and organism pages back to BRC Analytics**.  
- This would allow users browsing NCBI datasets to directly discover relevant resources and analytics available in BRC Analytics.  

### Impact  
- Improves discoverability of BRC Analytics resources from widely used NCBI entry points.  
- Strengthens interoperability between NCBI and BRC Analytics.  
- Provides researchers with a smoother navigation experience across both platforms.  

### High-Level Tech Design

1. NCBI will periodically call BRC Analytics APIs to get the links to our [organism detail pages](https://brc-analytics.dev.clevercanary.com/data/organisms/73230) and our[ analyze pages](https://brc-analytics.dev.clevercanary.com/data/assemblies/GCA_001008285_1) for assemblies.
1. The API will be implemented by placing the JSON file at the appropriate path in the NextJS `public` folder during build time and serving the files statically from their path within public.  For example, a file at `public/api/assets/myfile.json` will be available after build/export at `/api/assets/myfile.json`.
1. At build time or pre-build step, the assemblies.json and organisms.json files will be processed to extract the relevant information for linking to BRC.
1. Links will be available at /api/links/organism-links.json and /api/links/assembly-links.json
1. Exact file format and URL path TBD.

### Subtasks

- [ ] UI mocks for NCBI site for reciprocal linking
- [ ] Download Assemblies API 
- [ ] Download Organisms API
- [ ] Organism page updates to add context to support arriving from an external site
- [ ] Analyze page updates to add context to support arriving from an external site
- [ ] Track inbound links from NCBI

d-callan:
"Exact file format and URL path TBD"

[...]

PR #893, opened 2025-10-06
feat: add NCBI cross-linking API endpoints

dannon:
Extends the backend to do something actually useful -- providing an API for external consumers

## Summary
- API endpoints providing linkout data for NCBI cross-referencing
- Organism links endpoint mapping taxonomy IDs to BRC Analytics pages
- Assembly links endpoint mapping accessions to BRC Analytics pages
- Cached responses with 24-hour TTL

**Depends on:** #892

### DIFF
--- a/.github/workflows/run-checks.yml
+++ b/.github/workflows/run-checks.yml
@@ -72,3 +72,51 @@ jobs:
       - name: poetry lint
         # Validate the GA2 catalog source files.
         run: poetry check -P catalog/py_package/
+
+  api-tests:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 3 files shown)


d-callan (approved):
thanks for this @dannon. its a good foundation to start from and id like to get it in so i can iterate on it soon. weve deviated from the spec in #783 a bit here, but since this isnt deployed anywhere yet and the requirements are changing a bit anyhow i dont think we should stop this for that.

related, since i mentioned changing reqs, ill update #783 soon w new info on what ncbi would like in addition to what was already proposed. any feedback there appreciated.

PR #904, opened 2025-10-10
feat: add ga2 bucket file

Smeds:
feat: add reference assembly files (2bit, fa and repeatModeler.families) located on genomeark bucket to source folder

## Description

Brief description of the changes made in this PR.

## Related Issue

If this addresses an existing GitHub issue, link it here (e.g., "Closes #123"). If no existing issue exists, consider creating one first to discuss the changes.


d-callan:
i think most of the assets this tries to pick up we have already supported via gxy data cache. the only one not covered by that case has unclear priority currently. im glad this pr exists, instructionally, but im going to close for now and use it as a reference only probably when we get to picking up ga2 assets from aws.

Issue #979, opened 2025-11-05
Update measles white paper to use Vega-Lite interactive images where possible


NoopDog:
# Overview
Some of the images in the [measles white paper](https://brc-analytics.dev.clevercanary.com/learn/featured-analyses/evolutionary-dynamics-of-coding-overlaps-in-measles) have interactive Vega-Lite versions. We need to:

1. Add support for vega-lite images
2. Update the images in the paper to the corresponding Vega-Lite images.

PR #980, opened 2025-11-05
feat: add vega embed component and use for measles whitepaper

dannon:
Implements https://github.com/galaxyproject/brc-analytics/issues/979

Need to make one cleanup pass, and we may want to adjust the vega spec to fit the expected dimensions better (I can elaborate), but this gets the job done:

image


NoopDog:
Thx so much @dannon what needs to be done to get this out of draft?

dannon:
Following up from the call -- primary followup would be adjusting the vega spec to set the initial coordinates and width a bit better, and potentially adjusting the legend (or clipping it), but neither of those would be barriers to merging this and iterating.

Issue #824, opened 2025-09-11
Custom workflow (upload data only)


NoopDog:
- [ ] Update the no workflow stepper to send to the data landing request
- [ ] Mock for selecting UCSC browser additional data
- [ ] Implement UCSC additional data selection

First pass data landing request shape:

- assembly
- gtf
- tracks
- read_runs

PR #993, opened 2025-11-10
fix: set fastq collection names (#824)

mvdbeek:
## Description

Fixes the `undefined` collection name

## Related Issue

Should be the last bit for #824 ?


mvdbeek:
I'd test it but it's been building since forever ...

NoopDog:
Sorry we missed this but will take yor names @mvdbeek  cc: @hunterckx

Issue #1047, opened 2025-12-17
External identifier drift, data provenance, and API snapshot clarity


d-callan:
## Description

We are increasingly relying on and exposing NCBI-derived data via JSON and a REST API, while tightly attaching records to external identifiers that 'change' over time.

We have already seen concrete examples of this causing confusion and maintenance burden (e.g. NCBI taxon ID and naming changes, superseded assemblies; see #499). As we ingest more data and expose it programmatically, this problem will grow.

## Core problem

- External identifiers are not stable
    - NCBI taxon IDs, assembly versions, and SRA metadata can change upstream.
- Our data is effectively a rolling snapshot
    - Each ingest overwrites prior JSON outputs.
    - API responses do not clearly indicate which upstream NCBI state they reflect.
- Identifiers are exposed as if they were truly stable
    - Users may assume reproducibility or one-to-one correspondence with NCBI.
- Caching amplifies the issue
    - API responses are cached (e.g. via Redis) using external identifiers as keys.
    - Upstream changes + cached responses can lead to stale or inconsistent results.

## Impact/ Urgency

Do I think this is 'critical' yet? Not quite, but I do think we should get ahead of it before we make it worse. Right now I'm pretty sure our worst case is something like an NCBI taxon id is cached in Redis as one we support, and then a change at NCBI means we have swapped it out for a different taxon id and don't really support it any longer. ideally what I think I'm looking for is explicit data snapshots with dates, which we can be transparent about with users. **Possibly for now we just disable caching until we have a chance to discuss?**

d-callan:
also possible implications to internal inconsistency if we have different scripts ingesting data at different times

d-callan:
Thanks. If that's our strategy for now, or really either way I guess, it seems important to make sure deployment process is documented. If it is somewhere already can you point me to it? I need to know exactly when to expect this thing to restart/ how to force it to..

PR #1048, opened 2025-12-18
fix: flush redis cache on API service startup

dannon:
## Summary
- Flushes all Redis cache on API service startup to avoid stale/inconsistent responses after catalog updates
- Adds proper lifespan handler to close Redis connection on shutdown

Partial mitigation for #1047.

## Test plan
- [x] Verified cache flush works on restart (set test key, restart, key is cleared)
- [x] Verified API health and cache health endpoints still work
- [x] Confirmed no external API endpoint exposes the flush functionality

### DIFF
--- a/backend/api/app/core/dependencies.py
+++ b/backend/api/app/core/dependencies.py
@@ -14,3 +14,9 @@ async def get_cache_service() -> CacheService:
         settings = get_settings()
         _cache_service = CacheService(settings.REDIS_URL)
     return _cache_service
+
+
+def reset_cache_service() -> None:
+    """Reset the global cache service instance (used during shutdown)"""
+    global _cache_service
+    _cache_service = None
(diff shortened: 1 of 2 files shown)


dannon:
@d-callan Realistically the reload scenario isn't possible with how we deploy (container restarts = fresh process), but it's a good catch -- I pushed a fix that resets `_cache_service = None` after closing. It's a cheap safeguard and makes the code more technically correct, which is the best kind of correct.

Re: the Copilot suggestions:
- **Error handling**: Intentionally not adding try/except. If Redis is unavailable on startup, we'd rather fail fast so the issue gets noticed and fixed, rather than silently limping along without caching and potentially causing harder-to-debug problems downstream.
- **Tests**: These are trivial wrappers - flush_all() is a one-liner around redis.flushdb(). If caching logic becomes more complex we can add integration tests, but not worth it for this.
- **"Don't close the connection"**: Bad suggestion - we should close cleanly, just need to reset the singleton afterward (which is now done).

d-callan (approved):

PR #1117, opened 2026-01-30
fix: ensure release workflow triggers deploy and uses correct version tag

dannon:
## Summary

Two issues discovered during the v0.20.0 release:

1. **Deploy workflow didn't auto-trigger**: `GITHUB_TOKEN` pushes don't trigger other workflows (GitHub prevents this to avoid infinite loops). Added explicit `gh workflow run` call to trigger the deploy.

2. **Version showed as `v0.20.0-1-gXXX` instead of `v0.20.0`**: Release Drafter creates tags pointing to commits on main, but after merging main→production there's a new merge commit. The tag was behind production HEAD, so `git describe` included the commit offset. Now we move the tag to point to the production merge commit before pushing.

## Changes

- Added step to move the release tag to the production merge commit
- Added step to explicitly trigger the deploy workflow after pushing
- Added comments explaining why these steps are needed

## Test plan

- Reflect on this next release and update as necessary.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,7 +45,6 @@ backend/**/.env
 # python
 venv
 __pycache__
-
 playwright-report/
 
 # public api
--- a/.prettierignore
+++ b/.prettierignore
@@ -21,3 +21,6 @@ venv
 /catalog/output
 /catalog/ga2/output
 /catalog/schema/generated
+
+# claude code config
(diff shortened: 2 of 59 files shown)


dannon:
Since this only touches the release workflow I'm going to go ahead and merge this.

PR #1142, opened 2026-02-11
feat: add hi-c workflows

d-callan:
## Description

adding hi-c workflows under regulation workflow category. browsing iwc, and seems to me theyre the only relevant ones that are straightforward to add that we dont already have.


mvdbeek (approved):
LGTM, thank you @d-callan!

Issue #1128, opened 2026-02-05
Keycloak proof of concept for auth flow testing


NoopDog:
## Summary

Set up a Keycloak proof of concept for auth flow testing by adding Keycloak to the Docker Compose file, enabling developers to run BRC + Galaxy + auth locally as a full stack.

## Context

Dannon suggested using a local Docker container to run Keycloak for testing login access and auth flow improvements. The approach is to add it to Docker Compose so the full stack can be run locally.

## Theme

Build a reliable, scalable foundation

## Area

Galaxy Integration

Issue #812, opened 2025-09-08
Login/Access to users Galaxy API


MillenniumFalconMechanic:
# Overview

For tighter integration between Galaxy and the BRC Analytics site, it will be helpful for BRC to:

1. Know the user's identity
2. Know the user's associated Galaxy identity.
3. Perform some actions on behalf of the user in Galaxy.

This will enable the following use cases, for example:

1. Listing user histories
2. Selecting among existing histories to send data to.
3. Listing status of workflow runs
4. Maintaining a user profile with preferences and history.
5. Logging the user's BRC activity 
6. Allowing BRC to access / organize the user's workflow results

# Possible Approaches 

## Approach 1 - Token-exchange via OIDC/signed id_token

There are several ways to approach single sign-on with Galaxy. This approach aims to be low-effort on the Galaxy side while still securely meeting the requirements. 

If Galaxy can verify Google-issued id_token (OIDC), Galaxy can accept a proof that the user owns a Google account and mint a Galaxy API token for that identity or join that identity to an existing  Galaxy account. 

### High-Level Flow

1. User logs into BRC analytics, NextAuth gets the Google id_token or access_token.
2. User selects "Connect with Galaxy".
3. BRC calls Galaxy endpoint /api/exchange_oidc and presents the Google id_token (signed JWT).
4. Galaxy verifies the id_token signature and email_verified with Google (or via Google's jwks), then:
    1. if Galaxy account exists for that email, link it; else create a new account (optionally sending a welcome/verification email).
    2. Galaxy issues a Galaxy API token scoped for BRC (optionally a token with an indication it was issued via OIDC)
5. Galaxy returns the token to BRC; BRC stores it encrypted at rest and uses it on behalf of the user.
6. The user can disconnect the BRC from the Galaxy at any time.

#### Infrastructure Requirements
We would need a few items in place to implement this, including:

1. BRC Analytics NextJS server to run the NextAuth Google login.
2. A database to save the user's profile and API keys securely encrypted at rest.
3. The Galaxy endpoint to exchange the Google id_token
4. Galaxy to implement scoped API keys.

## Approach - 2 Let the user share their Galaxy API token with BRC.

 1. User logs into BRC analytics, NextAuth gets the Google id_token or access_token.
 2. User selects "Connect your Galaxy" account.
 3. User sees instructions on how to obtain a Galaxy API key.
 4. User goes to Galaxy, copies the API key and pastes it onto BRC Analytics 
[...]

PR #1152, opened 2026-02-17
feat: add Keycloak SSO testbed with BFF pattern

dannon:
## Summary

Adds Keycloak-based SSO authentication for BRC Analytics using the Backend-for-Frontend (BFF) pattern. Tokens stay server-side in Redis; the browser only gets an opaque httpOnly session cookie. Includes a Docker Compose testbed with Keycloak, Galaxy, and a demo script showing end-to-end SSO flow.

This implements Milestone 1 from #812 — Galaxy running in the compose stack with shared Keycloak SSO and bearer token API access.

Closes https://github.com/galaxyproject/brc-analytics/issues/1128

### What's here

**Backend (FastAPI):**
- OIDC auth endpoints: `/auth/login`, `/auth/callback`, `/auth/me`, `/auth/logout`
- PKCE (S256) for the authorization code flow
- Redis-backed session store with token refresh
- Keycloak config via environment variables

**Frontend:**
- `BrcAuthProvider` context + `useAuth()` hook
- `AuthButton` component in the header (Sign In / user name + Sign Out)
- Gracefully degrades when backend URL is unset (static export still works)

**Infrastructure:**
- `docker-compose.auth.yml` with Keycloak, PostgreSQL, Galaxy, Redis, backend
- Pre-configured Keycloak realm with `brc-analytics` and `galaxy-api` clients
- `keycloak-init` sidecar that creates audience mapper + Galaxy API scope via kcadm
- Galaxy OIDC config files for Keycloak integration
- `demo-sso.sh` exercising the full SSO flow end-to-end

### How to try it

```bash
# One-time: add keycloak to hosts (for split-horizon DNS)
sudo sh -c 'echo "127.0.0.1 keycloak" >> /etc/hosts'

# Start the stack (Galaxy takes a few minutes on first boot)
docker compose -f docker-compose.auth.yml up --build

# In another terminal, start the frontend
npm run dev

# Run the demo
./demo-sso.sh
```

Test user: `testuser` / `testpass`

## Test plan

- [x] `npm run check-format && npm run lint && npx tsc --noEmit` all pass
- [x] `npm run test` — all 387 tests pass
- [ ] `docker compose -f docker-compose.auth.yml up` starts cleanly
- [ ] Login flow: frontend Sign In → Keycloak → redirect back → user name shown
- [ ] Logout clears session
- [ ] SSO: log into Galaxy via Keycloak, BRC Analytics auto-authenticates
- [ ] Bearer token: `demo-sso.sh` calls Galaxy API with Keycloak token

### DIFF
--- a/backend/api/app/api/v1/auth.py
+++ b/backend/api/app/api/v1/auth.py
@@ -123,7 +123,7 @@ async def logout(
 ) -> JSONResponse:
     """Clear the user's session from Redis and remove the cookie."""
     if brc_session:
-        await auth.delete_session(brc_session)
+        await auth.revoke_session_tokens(brc_session)
 
     resp = JSONResponse(content={"message": "Logged out"})
     resp.delete_cookie(COOKIE_NAME, path="/")
--- a/backend/api/app/services/auth_service.py
+++ b/backend/api/app/services/auth_service.py
@@ -235,9 +235,15 @@ class AuthService:
                 )
                 claims = self.decode_token_claims(refreshed["access_token"])
             else:
-                # Refresh failed, session is effectively dead
-                await self.delete_session(session_id)
-                return None
+                # Refresh failed — possibly because a concurrent request
+                # already used the refresh token (Keycloak rotates them).
+                # Re-read the session; if it was updated, use the new tokens.
+                session = await self.get_session(session_id)
+                if not session:
+                    return None
+                claims = self.decode_token_claims(session["access_token"])
+                if claims.get("exp", 0) < time.time():
+                    return None
 
         return {
             "sub": claims.get("sub"),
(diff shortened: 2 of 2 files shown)


dannon:
Yep, driven by the site config.  Without auth configured, none of it should show up, but I'll double check.

As far as it being draft, I guess we could go ahead and push forward without an auth server to target, but I'd hesitate to build more on top of it until that's seriously underway.  I'll see if we can get an estimate for that.

dannon (commented):
Reviewed the suggestions. Three are worth acting on: the `useEffect` in `authentication.tsx` should skip the `/auth/me` fetch when `loginEnabled` is false, the `AuthButton` registration in `config.ts` should be conditional rather than relying on the component returning null, and the auth service needs a `close()` call in the lifespan shutdown in `main.py`. Will fix those.

The remaining six are all accurate observations in isolation but don't apply here. The hardcoded client secret default, `bruteForceProtected`/`sslRequired` settings, `VERIFY_SSL=False`, and the demo script's trap cleanup are all intentional choices for a local Docker testbed that gets torn down regularly — none of these files are used in production. The JAR checksum pinning and OIDC config TTL are reasonable ideas in general but premature for a dev testbed that's still taking shape.

dannon:
I'll rebase this to resolve conflicts, thank you for the approve!

PR #1187, opened 2026-03-11
feat: filter new workflows page to only those w valid assemblies available

### LINKED ISSUE #1186: Validate workflow–assembly compatibility for workflow-first UI
## Problem

The site now supports two navigation patterns:

- Assembly → Workflow (existing)
- Workflow → Assembly (new “Workflows” page)

Workflow–assembly compatibility is currently determined at runtime by filtering workflows against assembly metadata (taxon IDs, ploidy, etc.). This logic works reasonably well for the assembly-first route.

However, in the workflow-first view we can list workflows that ultimately have **no valid assemblies**. 

## Concern

We should not simply hide these workflows at runtime, because a workflow having zero valid assemblies may indicate a configuration or annotation error (e.g., incorrect taxon ID). If filtered silently, these problems may go unnoticed.

## Proposed Direction

Add validation to ensure workflows listed in the workflow-first UI have at least one compatible assembly.

Workflows that resolve to zero assemblies should be detected and reported during the build or CI process so that mismatches in workflow annotations are visible and can be fixed.

d-callan:
## Description

the ts bit of the catalog build now introduces a step which consumes workflows and assemblies json, produces a mappings json + qc report. the ui uses the mappings json to decide which workflows to show in workflows page. hopefully a human uses the qc report to make sure the mappings json is sensible.

only 13 in ga2 now:
image

## Related Issue

closes #1186

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "brc-analytics",
-  "version": "0.23.0",
+  "version": "0.24.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "brc-analytics",
-      "version": "0.23.0",
+      "version": "0.24.0",
       "dependencies": {
         "@databiosphere/findable-ui": "^48",
         "@emotion/react": "^11",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "brc-analytics",
-  "version": "0.23.0",
+  "version": "0.24.0",
   "private": true,
   "scripts": {
     "dev": "./scripts/dev.sh brc-analytics && ./scripts/set-version.sh dev && ./scripts/sync-api-brc-analytics.sh && next dev",


dannon:
Copilot's review is mostly reasonable here — the bundle size concern is probably the most important one. The UI only needs to know "does this workflow have any compatible assemblies?" but you're shipping the full accession lists to the client (68K lines for BRC). A smaller artifact with just the trsId set or a count would keep the client lean. The duplicate workflow entries in the QC report are also worth a look — seems like the same workflow appears in multiple categories and the flatten doesn't dedup, so you get repeated rows. The stale `.d.ts` cleanup in the sync scripts and the missing `default` in the switch are both easy wins. The `Object.keys(mappings).length` vs `mappings.length` thing is obviously just a typo since GA2 gets it right.

A few things Copilot didn't mention:

The compatibility logic in `build-workflow-mappings.ts` is a straight port of `workflowPloidyMatchesOrganismPloidy` from `app/apis/catalog/brc-analytics-catalog/common/utils.ts` and `workflowIsCompatibleWithAssembly` from `app/components/Entity/components/AnalysisMethodsCatalog/utils.ts`. Having two copies means they need to stay in sync — might be worth importing the originals (or extracting a shared module) rather than maintaining parallel implementations. The `AssemblyForCompatibility` type constraint you're using is already a subset of the app types so this should be doable with generics.

The GA2 build reads `catalog/output/workflows.json` (a BRC build artifact), which creates an implicit ordering dependency — BRC must always be built before GA2. That's probably fine in practice but worth a comment in the code so someone doesn't rearrange the build steps and get confused.

PR #1222, opened 2026-04-03
feat: stub add flu workflow

d-callan:
## Description

Activates the influenza consensus sequence and subtyping workflow, enabling direct launch in Galaxy with a basic "Launch in Galaxy" button. This PR does NOT include workflow stepper UI yet.

**Changes**

- Workflow Activation: Enabled flu workflow in YAML with ORGANISM scope and taxonomy ID 11320
- Scope-based Architecture: Refactored ConfiguredValue types to support ASSEMBLY, ORGANISM, and SEQUENCE scopes
- Feature Flag: Added flu workflow behind feature flag 'flu' for controlled rollout
- Type Safety: Introduced proper type guards and discriminant properties for ConfiguredValue union types


app/views/WorkflowsView/utils.ts
```diff
@@ -83,6 +86,14 @@ export function getWorkflows(
   for (const category of workflowCategories) {
     if (!category.workflows) continue;
     for (const workflow of category.workflows) {
+      // Skip flu workflow - it's handled separately with feature flag below.
```
hunterckx: For the sake of consistency with how other YAML-sourced workflows are handled, and to consolidate code, it seems to me that it would be best to just check the feature flag here and _conditionally_ skip the flu workflow (and remove the conditional further down that inserts it separately)

app/components/Entity/components/ConfigureWorkflowInputs/components/Main/components/Stepper/components/Step/hooks/UseLaunchGalaxy/useLaunchGalaxy.ts
```diff
@@ -11,9 +11,115 @@ import { CUSTOM_WORKFLOW } from "../../../../../../../../../../../../views/Analy
...
+ * @param configuredValue - Configured values for the workflow.
+ * @param origin - Origin URL for the Galaxy instance.
+ * @param run - Async runner function from useAsync hook.
+ * @returns Promise resolving to the Galaxy workflow landing URL.
+ */
+async function getLandingUrlForWorkflow(
+  workflow: Workflow,
+  configuredValue: ConfiguredValue,
+  origin: string,
+  run: (promise: Promise<string>) => Promise<string>
```
hunterckx: It might be best to remove `run` from this function and instead wrap it around the call to `getLandingUrlForWorkflow` -- that would simplify this function and would keep `run` close to the hook that it relates to

hunterckx (changes requested):
A couple things that I think would be good to adjust before merging!

hunterckx (approved):
Thanks @d-callan!

PR #1227, opened 2026-04-11
fix(backend): chain FastMCP lifespan into FastAPI app

dannon:
## Summary

- `/api/v1/mcp/` has been returning 500 on every request because the mounted FastMCP `http_app`'s lifespan was never entered, so its `StreamableHTTPSessionManager` task group wasn't running. The fix lifts `mcp` / `mcp_app` construction to module scope in `backend/api/app/main.py` and wraps the parent FastAPI lifespan in `async with mcp_app.lifespan(app):`, which is the pattern FastMCP docs suggest for mounting into a FastAPI app.
- Adds `backend/api/tests/test_mcp.py` with `TestClient`-based coverage of the MCP surface: an `initialize` round-trip, a `tools/list` check that all expected tools are registered, and a `tools/call` against `search_organisms` backed by the same sample catalog fixture used in `test_catalog_data.py`. I confirmed the tests fail on the pre-fix code with the original `RuntimeError: Task group is not initialized` and pass once the fix is applied.

## Test plan

- [x] `pytest backend/api/tests/test_mcp.py` passes (3/3)
- [x] Local `curl` + MCP Inspector CLI round-trip against `uvicorn app.main:app`
- [ ] Re-verify against `https://dev.brc-analytics.org/api/v1/mcp/` after dev deploy

### DIFF
--- a/backend/api/app/main.py
+++ b/backend/api/app/main.py
@@ -28,36 +28,36 @@ if settings.SENTRY_DSN:
         traces_sample_rate=1.0,
     )
 
+# Build the MCP ASGI app at module scope so its lifespan can be chained into
+# the parent FastAPI lifespan below. Mounting a FastMCP http_app without
+# running its lifespan leaves the StreamableHTTPSessionManager task group
+# uninitialized, which causes every MCP request to 500.
+mcp = create_mcp_server(get_catalog_data(), get_ena_service())
+mcp_app = mcp.http_app(path="/", stateless_http=True)
+
 
 @asynccontextmanager
 async def lifespan(app: FastAPI):
     """Startup and shutdown events for the application"""
-    # Startup: initialize services and flush cache
-    cache_service = get_cache_service()
-    await cache_service.flush_all()
-    logger.info("Cache cleared on startup")
+    async with mcp_app.lifespan(app):
+        # Startup: initialize services and flush cache
+        cache_service = get_cache_service()
+        await cache_service.flush_all()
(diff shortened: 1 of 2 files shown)


backend/api/app/main.py
```diff
@@ -28,36 +28,36 @@
         traces_sample_rate=1.0,
     )
 
+# Build the MCP ASGI app at module scope so its lifespan can be chained into
+# the parent FastAPI lifespan below. Mounting a FastMCP http_app without
+# running its lifespan leaves the StreamableHTTPSessionManager task group
+# uninitialized, which causes every MCP request to 500.
+mcp = create_mcp_server(get_catalog_data(), get_ena_service())
```
mvdbeek, line 35: Is there no factory pattern possible ? on the galaxy side we use `python -m uvicorn --app-dir lib --factory galaxy.webapps.galaxy.fast_factory:factory` and the upside is that an import won't start the whole thing ...

mvdbeek (approved):

backend/api/app/main.py
dannon: Yep, that'd be cleaner. Galaxy already had the factory pattern in place though -- adding one here would be a bigger change since it touches the `Dockerfile CMD`, `docker-compose`, dev scripts, etc. Kept this one intentionally tight to just fix the immediate 500s following the documented FastMCP mounting pattern. Good follow-up though.  Thanks for the approve, I'll get this in and look at the refactoring.

PR #1260, opened 2026-04-29
feat: adding a "uploading your own data" help doc

scottcain:
## Description

This document tries (pretty successfully I think) to be a comprehensive answer to the question "How do I get my data into Galaxy if I said I would upload my own?" in the workflow stepper. It would probably be good after this is merged to add a link to this document from the stepper.

## Related Issue

I don't think this directly addresses an issue--it's just something I think was needed.

## Implementation notes

I put all of the image files in the same directory--I would guess that they might have to go somewhere else, in which case, we'd need to update the url references to the images.  Accordingly, I'm going to make this a draft until that's clarified.


NoopDog:
Hi @scottcain, Thanks for preparing this! How can we help get this out of draft?
Cheers,  and Thanks!

cc: @frano-m @d-callan

NoopDog (approved):

PR #1262, opened 2026-05-01
feat(evals): pydantic-evals harness for AI-enabled services

dannon:
## Summary

Adds an evals harness under `backend/api/evals/` for the inference-enabled
services -- search query interpretation, MCP tool selection, workflow
recommendation, and the multi-turn assistant. Built on `pydantic-evals` and
patterned after Galaxy's agent-evals-harness.

The harness is a thin runner: each dataset module declares its cases and
evaluators, `tasks.py` adapts our existing `LLMService` / `AssistantAgent`
into pydantic-evals tasks (no service code is touched), `model_registry.py`
loads candidate models from `models.yaml`, and `run_evals.py` produces a
markdown report per run. Free models (gpt-oss, Llama-3.3, Maverick) and the
Anthropic line are already wired up; results from a few baseline runs are
checked into `evals/results/`.

A couple of things worth flagging:

- Cache is rebuilt per (dataset, model) iteration so `LLMService`'s
  content-keyed cache entries can't leak from one model's output into
  another model's input. Multi-turn sessions still work because session ids
  are minted per case.
- Structural failures only count toward denominators of evaluators that
  were actually declared on that case, including pydantic-evals' duplicate
  suffixing (`FieldEquals_2`) and LLMJudge's `_score`/`_pass` split.
- `models.yaml` is gitignored; `models.yaml.sample` is the template.

## Test Plan

- [ ] `uv run pytest evals/tests` (36 tests) passes
- [ ] `uv run ruff check evals` clean
- [ ] Spot-check a smoke run: `python -m evals.run_evals --datasets assistant_multiturn,search_interpretation --models claude-sonnet-4-6 --only exploration_only,gibberish` produces a report under `evals/results/`
- [ ] Verify the report renders correctly (per-case cells, partial-coverage denominators, failure rows)

### DIFF
--- a/backend/api/app/core/cache.py
+++ b/backend/api/app/core/cache.py
@@ -109,7 +109,9 @@ class CacheService:
         """Generate a cache key from prefix and parameters"""
         # Sort parameters for consistent keys
         param_str = json.dumps(params, sort_keys=True, default=str)
-        hash_val = hashlib.md5(param_str.encode()).hexdigest()[:16]
+        hash_val = hashlib.md5(
+            param_str.encode(), usedforsecurity=False
+        ).hexdigest()[:16]
         return f"{prefix}:{hash_val}"
 
     async def close(self):
--- a/backend/api/evals/.gitignore
+++ b/backend/api/evals/.gitignore
@@ -1,0 +1,4 @@
+models.yaml
+*.local.md
+__pycache__/
+.pytest_cache/
--- a/backend/api/evals/README.md
+++ b/backend/api/evals/README.md
@@ -19,7 +19,12 @@ Mirrors the Galaxy `agent-evals-harness` shape: a thin runner around
    (or `api_key_env` references). The `judge:` key must point at one of the
    declared models.
 2. `uv sync --extra dev` from `backend/api/` to pull in `pydantic-evals`.
-3. `export ANTHROPIC_API_KEY=...` (or whatever `api_key_env` you used).
+3. Export whichever env vars your `models.yaml` references via `api_key_env`
+   (e.g. `ANTHROPIC_API_KEY`, `OPENAI_API_KEY`). Models with a literal `api_key`
+   in YAML don't need an env var. The harness stubs the unused
+   `ANTHROPIC_API_KEY`/`OPENAI_API_KEY` so backend init (which reads
+   `AI_API_KEY` / `OPENAI_API_KEY` from `Settings`) doesn't crash on startup --
+   `_override_model` swaps in the real model immediately after.
 
 ## Running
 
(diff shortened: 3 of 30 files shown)


dannon:
I’m going to trim/consolidate some of the checked-in eval results before moving this out of draft. We want a useful running benchmark that can accumulate over time, but the current report set is a bit noisy and hard to scan.

dannon:
Triaged the 7 inline comments. One real bug, four defense-in-depth, one already fixed but unpushed, one stale-artifact misattribution.

**Real bug**
- `_fmt_score` numerator (`report.py:62`) -- `int(round(s*n))/{n}` silently lied for fractional evaluators; a keyword-fraction or continuous judge score got rounded back to a fake integer pass-count. Dropped the rounding, format the numerator with one decimal: binary cases still read cleanly (`5.0/5 (1.00)`), fractional ones stay honest (`3.5/5 (0.70)`). `c661b6b9`.

**Defense-in-depth (not bugs in current code, but free to harden)**
- `ctx.metadata.get(...)` in `IwcIdInSet` / `FinalSchemaContains` -- the metadata branch is only the fallback when the evaluator wasn't given args, and every case in `_CASES` passes args explicitly. Unreachable today, but `meta = ctx.metadata or {}` is one line and makes the failure mode explicit. `b46ee196`.
- `LLMJudge.score` / `.assertion` `.get(...)` -- pydantic-evals lets these be `bool | dict`. Our judges always use the dict form, so the `.get()` is fine in practice; `score=True` would crash. Added `isinstance(..., dict)` guard with default-name fallback. `62253999`.
- `hashlib.md5` in the eval cache and `app/core/cache.py` -- not security-sensitive (per-run content hash), but bare `md5()` blows up on FIPS-enabled Python. Tagged both with `usedforsecurity=False`. `252fd235`.

**Already fixed (just unpushed)**
- README "no-op stub" was fixed locally in `e181b4ec` on May 1 but never pushed -- that's on me, not on the review. Pushing this branch now surfaces it. README at HEAD describes the cache as a per-run in-memory dict rebuilt via `with_fresh_cache()`.

**Not a code bug -- stale artifact**
- Duplicate `covid_consensus` / `flu_consensus` rows in `2026-05-01-workflow-rec-fixed.md` -- the renderer at HEAD already dedupes per-case rows via a set union (`{c.name for ...} | {n for ...}`). That artifact was generated by an earlier renderer that concatenated two sorted sets, so a case that succeeded for some models and failed for others appeared in both halves. Nothing to fix in code; the file will regenerate cleanly next time the harness runs.

Copilot review: solid catch on the `_fmt_score` rounding, generic on the rest. 4/10.

NoopDog (approved):

PR #1212, opened 2026-03-24
feat: analysis assistant stepper handoff and many UX improvements

dannon:
## Summary

A batch of UX and agent behavior improvements to the analysis assistant, plus the integration point connecting a completed assistant session to the workflow configuration stepper.

### Agent behavior (backend)
- **Schema context injection** -- each user message now includes the current analysis state so the LLM knows what's been decided and doesn't re-ask about filled fields
- **Handoff URL reliability** -- fallback catalog lookups for assembly accession (by strain name, preferring reference) and workflow trsId (by name) when the LLM doesn't include them in the expected format
- **Field clearing** -- the LLM can now emit `null` in SCHEMA_UPDATE to clear downstream fields when the user changes their mind (e.g. switching analysis types). System prompt teaches the dependency chain

### Session management
- **Session persistence** -- session ID stored in localStorage, restored on mount via new `GET /session/{id}` endpoint. Suggestions now persisted in session state
- **New Conversation button** -- clears localStorage and resets all state, with fire-and-forget `DELETE /session/{id}` cleanup
- **Error recovery** -- failed messages stay visible with a Retry button instead of being silently removed

### UI polish
- **Content centering** -- wrapped assistant layout in `SectionContent` using the shared `sectionLayout` mixin (max-width 1136px) to match Learn/About pages
- **BRC logo** -- replaced the blue "AI" circle avatar with the BRC Analytics logo
- **Suggestion chips** -- styled with light blue border, hover state, and better padding

### Stepper handoff
- When the user clicks "Continue to Workflow Setup", the assistant writes a handoff payload to localStorage with the data source preference (ENA vs upload)
- The stepper page reads and consumes this on mount, advances past the auto-configured assembly and GTF steps, and pre-selects the sequencing data toggle
- Stale handoffs (>30 min) are rejected; page refresh gives normal stepper behavior

### Trust & transparency

Two follow-ups from feedback on an earlier demo of this branch:

- **No third-party download instructions** -- new "When data is missing" guidance in the system prompt tells the agent not to send users to manually download GTFs from VEuPath/NCBI/Ensembl and re-upload (format conversion isn't trivial and the agent can't verify what's needed). Instead it acknowledges the gap and points to https://help.brc-analytics.org or the issue tracker
[...]

PR #1265, opened 2026-05-07
feat(assistant): prompt-injection hardening (audit follow-up to #1212)

dannon:
Follow-up to #1212. Tightens six attack surfaces in the assistant chat flow
surfaced by a prompt-injection audit of that branch.

> **Note:** this branch is built on top of #1212, so the diff includes those
> commits until #1212 merges. The new work is in commits `bd2df2e2`..`38f3ba29`.

## What changes

1. **System-prompt hardening** -- a dedicated section tells the model to treat
   role-override attempts (claiming to be a system prompt, "ignore the rules
   above", etc.) as off-topic and redirect.
2. **User-input fencing** -- user messages are wrapped in
   `<user_input>...</user_input>`. The system prompt instructs the model to
   treat fenced content as untrusted data. Literal closing tags in the user's
   message are neutralized with a zero-width space so the fence is
   unambiguous.
3. **Session-cookie binding** -- `POST /chat` sets an httpOnly Same-Site=Strict
   cookie containing an HMAC of the session_id; `GET`/`DELETE /session/{id}`
   verify the cookie. A leaked session_id alone (browser logs, shoulder-surf,
   etc.) is no longer enough to read or delete a session. Falls back to
   legacy unbound mode when `SESSION_COOKIE_SECRET` is empty so local dev
   keeps working.
4. **X-Forwarded-For trust scoping** -- the rate limiter only honors XFF when
   `TRUST_PROXY_HEADERS=true`. Default is now `false`; the previous code
   blindly trusted XFF, which lets clients spoof IPs and slip past per-IP
   rate limits when the app is exposed directly.
5. **\`state.messages\` cap** -- bounds user-facing chat history at 100
   messages so a long-lived session can't bloat Redis payloads.
6. **CORS wildcard rejection** -- \`Settings.__init__\` raises if
   \`CORS_ORIGINS=*\` and \`ENVIRONMENT\` is not local/dev/development. With
   unauthenticated endpoints, wildcard CORS lets any site initiate chats
   from a victim's IP.

## New env vars (for ops)

- \`SESSION_COOKIE_SECRET\` -- required in prod. Generate with
  \`python -c "import secrets; print(secrets.token_urlsafe(48))"\`.
- \`TRUST_PROXY_HEADERS=true\` -- set this when deploying behind nginx in
  docker-compose, otherwise rate-limit keys will resolve to the nginx
  container IP instead of the real client.

Both are documented in \`backend/api/.env.example\`.

## Tests

100 passing (75 baseline + 25 new). New coverage:
- \`test_session_signing.py\` -- HMAC sign/verify round-trip + tamper cases
- \`test_assistant_endpoints.py\` -- 6 integration tests covering the cookie
  contract end-to-end via TestClient
- \`test_rate_limit.py\` -- 4 tests for the XFF trust gate
- \`test_config.py\` -- 5 tests for CORS wildcard rejection
- \`test_assistant_agent.py\` -- new classes for system-prompt content,
  \`<user_input>\` fencing (including the fake-closing-tag case), and the
  message cap

## Test plan
- [ ] CI passes
- [ ] After #1212 lands, rebase onto upstream main and confirm diff is just
      the new commits

### DIFF
--- a/app/components/Assistant/SchemaPanel/schemaPanel.tsx
+++ b/app/components/Assistant/SchemaPanel/schemaPanel.tsx
@@ -44,11 +44,13 @@ const FIELD_ORDER: (keyof AnalysisSchema)[] = [
   ...OPTIONAL_FIELDS,
 ];
 
-const EMPTY_FIELD: SchemaFieldState = {
+// Frozen so the shared reference across PLACEHOLDER_SCHEMA fields can't
+// be mutated through any one key.
+const EMPTY_FIELD: SchemaFieldState = Object.freeze({
   detail: null,
   status: "empty",
   value: null,
-};
+});
 
 const PLACEHOLDER_SCHEMA: AnalysisSchema = {
   analysis_type: EMPTY_FIELD,
--- a/app/hooks/useAssistantChat.ts
+++ b/app/hooks/useAssistantChat.ts
@@ -1,3 +1,4 @@
+import { HTTPError } from "ky";
 import { useCallback, useEffect, useRef, useState } from "react";
 import { llmAPIClient } from "../services/llm-api-client";
 import {
@@ -165,12 +166,14 @@ export const useAssistantChat = (): UseAssistantChatReturn => {
  * @returns A user-facing error string
  */
 function handleChatError(error: unknown): string {
-  const err = error as { message?: string; name?: string };
-  if (err.name === "TimeoutError" || err.message?.includes("504")) {
+  const status =
+    error instanceof HTTPError ? error.response?.status : undefined;
+  const name = (error as { name?: string }).name;
+  if (name === "TimeoutError" || status === 504) {
     return "The assistant took too long to respond. Please try again.";
-  } else if (err.message?.includes("503")) {
+  } else if (status === 503) {
     return "The analysis assistant is currently unavailable. Please try again later.";
-  } else if (err.message?.includes("429")) {
+  } else if (status === 429) {
     return "Too many requests. Please wait a moment and try again.";
   }
   return "Something went wrong. Please try again.";
--- a/app/services/llm-api-client.ts
+++ b/app/services/llm-api-client.ts
@@ -14,6 +14,12 @@ import {
 } from "../types/api";
 
 const apiClient = ky.create({
+  // Send the assistant session-binding cookie cross-origin. Without
+  // this, ky inherits fetch's "same-origin" default and the browser
+  // strips the cookie when NEXT_PUBLIC_BACKEND_URL points off-origin
+  // (which is the deployed shape) -- silently breaking session
+  // restore/delete. Backend already sets CORS allow_credentials=True.
+  credentials: "include",
   hooks: {
     beforeError: [
       (error): HTTPError => {
--- a/app/views/WorkflowInputsView/hooks/UseAssistantHandoff/useAssistantHandoff.ts
+++ b/app/views/WorkflowInputsView/hooks/UseAssistantHandoff/useAssistantHandoff.ts
@@ -28,7 +28,12 @@ function readAndConsumeHandoff(): AssistantHandoff | null {
       return cachedHandoff;
     }
     // No fresh handoff: fall back to whatever the previous read produced so
-    // the strict-mode remount sees the same value as the first mount.
+    // the strict-mode remount sees the same value as the first mount. Still
+    // honor MAX_AGE_MS -- the module cache survives navigations, so without
+    // re-checking we could replay an expired handoff later in the tab.
+    if (cachedHandoff && Date.now() - cachedHandoff.timestamp > MAX_AGE_MS) {
+      cachedHandoff = null;
+    }
     return cachedHandoff ?? null;
   } catch {
     cachedHandoff = null;
--- a/backend/api/.env.example
+++ b/backend/api/.env.example
@@ -41,6 +41,11 @@ ENVIRONMENT=development
 RATE_LIMIT_REQUESTS=100
 RATE_LIMIT_WINDOW=60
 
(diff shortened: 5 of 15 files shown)


dannon:
Thanks @copilot-pull-request-reviewer -- two real bugs in there, a couple of cleanups, and one miss.

**Real bugs, fixed:**
- **`cachedHandoff` TTL re-check**: the freshness check only ran on the localStorage read, not the module-cache fallback, so a stale handoff could replay later in the tab. Now revalidates the timestamp before returning the cached value. ([d6e1e1d](https://github.com/galaxyproject/brc-analytics/commit/d6e1e1db))
- **`wf.get("trsId", iwcId)`**: `_condense_workflow` stores `trsId: None` when unset, so the default arg never fired. Switched both call sites to `wf.get("trsId") or wf.get("iwcId")`. ([2d27fa8](https://github.com/galaxyproject/brc-analytics/commit/2d27fa8e))

**Cleanup, not bugs:**
- **Cookie `Secure` flag** -- the cookie value is an HMAC of the session_id, useful only if you already have the session_id from the same connection, at which point Secure isn't what's saving you, TLS is. Our prod is always HTTPS with no HTTP listener for the cookie to leak to. Flipped it on outside `local`/`dev`/`development` anyway since the conditional is free. ([bad225e](https://github.com/galaxyproject/brc-analytics/commit/bad225ef))
- **Cookie name set/read drift** -- `SESSION_COOKIE_NAME` is a hardcoded literal in `config.py`, not env-driven, and the set + read sides live in the same file. Drift would require someone making it configurable and missing the rename. Tightened it up with `Cookie(alias=...)` against that hypothetical anyway. Same commit.

**Not applicable** -- the "`initialActiveStepOverride` won't apply if steps load async" concern: `configuredSteps` is a `useMemo` over a synchronous `getWorkflow(trsId)`. Populated at mount, only changes on a route swap that remounts the component. No empty→populated transition to guard against.

Side note: five copies of the same cookie-naming comment on five adjacent lines was a bit much. Counted as one finding regardless.

**Copilot review: 2/10.**

dannon:
Round 2 thoughts, @copilot-pull-request-reviewer:

**Cookie naming, again:** four copies of the same comment, but this time the framing is correct -- my round-1 fix unified the read side and left `_set_session_cookie` writing via `settings.SESSION_COOKIE_NAME`. Both still resolve to the same lru_cached value so nothing was actually broken, but the "stays in lockstep" comment I added was only half-honored. Tightened in [`a6fdfce2`](https://github.com/galaxyproject/brc-analytics/commit/a6fdfce2).

**Genuine new ones, both worth it:**
- **Shared `EMPTY_FIELD` reference in `PLACEHOLDER_SCHEMA`** -- no code mutates it today, but `Object.freeze` is free defense-in-depth so any future accidental write throws instead of silently corrupting all seven fields. [`a220efd6`](https://github.com/galaxyproject/brc-analytics/commit/a220efd6).
- **`err.message?.includes("504")` in `handleChatError`** -- fair catch, substring match on error text is fragile. Switched to `HTTPError.response?.status` from ky so the mapping is keyed on structured data. [`b5bba930`](https://github.com/galaxyproject/brc-analytics/commit/b5bba930).

Side note: posting the cookie-name comment four times on four adjacent lines, after I already addressed it in round 1, is the kind of behavior that drags the signal-to-noise down. One comment with the corrected framing would have been enough.

**Copilot review, round 2: 3/10 (recycle credit dampens the score).**

dannon:
Round 3, @copilot-pull-request-reviewer:

All three applied, but for the third pass in a row the framing assumes a deployment that isn't ours.

- **`credentials: "include"` on the ky client** ([7be0990a](https://github.com/galaxyproject/brc-analytics/commit/7be0990a)) -- prod is a static export with a relative `API_BASE_URL` routed through a same-origin proxy, so cookies flow without it. The change only matters for local-dev testing of the binding flow with the secret set, which is a narrow case. Harmless to apply.
- **Raise on empty `SESSION_COOKIE_SECRET` in non-dev** ([efe0a569](https://github.com/galaxyproject/brc-analytics/commit/efe0a569)) -- the secret is set by our ansible playbook, so the "operator forgets the env var" failure mode can't occur here. Belt-and-suspenders, mirrors the CORS guard, fine.
- **Skip empty XFF entries** ([f10b7714](https://github.com/galaxyproject/brc-analytics/commit/f10b7714)) -- `TRUST_PROXY_HEADERS` is opt-in and the proxy in front is ours; a malformed XFF can't reach the parser. Hardened anyway because the change is four lines.

Side note: this is the third review in twelve hours where the bug-shaped framing assumes a different deployment than the one we actually run. We accept the cheap defensive changes, but "real bug" should mean "fires in our actual codebase and deployment" rather than "fires under some hypothetical operator misconfiguration."

**Copilot review, round 3: 0/10 on real bugs, 3/3 on defensive nits accepted.**

NoopDog:
Hi @dannon, can you resolve the comments for any fixes pushed or rejected feedback? Thanks!

dannon:
**Copilot review, round 5: 0/4 -- nothing actionable.**

Three of the open threads (`assistant.py:22`, `config.py:55`, `assistant_agent.py:393`) are already addressed in the rebase -- GitHub flags them outdated.

The one new comment claims `const EMPTY_FIELD: SchemaFieldState = Object.freeze({...})` won't compile under strict TS because `Object.freeze` returns `Readonly<T>`. Built a minimal repro on TS 5.9.3 (the version pinned here), `strict: true`, identical shape -- `tsc --noEmit` exits 0. `Readonly<T>` is structurally assignable to `T` (the `readonly` modifier is a write-protection annotation at the alias site, not a distinct type). CI's `run-checks` confirms -- green on `3c766b91`.

Resolving the threads. No code change.

NoopDog (approved):

PR #1280, opened 2026-05-20
feat: add some assembly building and polishing workflows

### LINKED ISSUE #705: add assembly workflows to organisms tab?
right now i dont find organisms tab super useful if im honest. just another way to get to an assembly.. but its occurred to me that this could serve as a natural home for assembly workflows. organisms table could gain a column w buttons at left similar to how assemblies table does.

d-callan:
## Description

activating some assembly building and polishing workflows. they arent quite ready for prod here, but only bc we still need organism scope workflow stepper fleshed out. theres no feature flag here, but there could be if we wanted.

## Related Issue

closes #705

### DIFF
--- a/app/views/AnalyzeWorkflowsView/components/Main/main.tsx
+++ b/app/views/AnalyzeWorkflowsView/components/Main/main.tsx
@@ -1,4 +1,5 @@
 import { BackPageContentMainColumn } from "@databiosphere/findable-ui/lib/components/Layout/components/BackPage/backPageView.styles";
+import { useFeatureFlag } from "@databiosphere/findable-ui/lib/hooks/useFeatureFlag/useFeatureFlag";
 import { JSX } from "react";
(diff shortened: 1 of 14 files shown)


hunterckx (commented):
A couple things I want to make sure of before approving:
1. Does the Shovill assembly deliberately have no parameters enabled? Seems like it's currently not using any information from BRC Analytics
2. Is the issue referenced in the PR description correct? It does look like it covers some relevant ideas but I'm having trouble discerning how directly it corresponds to what this PR is doing

Thanks @d-callan!

d-callan:
hi @hunterckx thanks for having a look. and yea the workflow inputs are a little funny i think, but suspect thats a thing to follow up w in the workflow. so maybe what ill do is add a feature flag so we can merge this while i keep figuring out input configuration.

hunterckx (approved):
Cool -- looks good to me!

PR #1286, opened 2026-05-27
ci: remove S3/CloudFront deploy workflows

dannon:
## Summary
- Delete `dev-deploy.yml` and `publish.yml` -- both synced builds to S3 and invalidated CloudFront, which we no longer use now that frontend and backend are hosted on TACC.
- Drop the "Trigger production deploy" step from `publish-release.yml` since `publish.yml` no longer exists; the TACC deploy is now manual after each release.

## Test plan
- [ ] Confirm `publish-release.yml` still parses (YAML lint via GH Actions on push)
- [ ] Next release cuts cleanly without trying to dispatch the removed workflow


dannon:
Thanks @copilot-pull-request-reviewer -- fair catch, that's exactly right. Removed in 3413809c.

NoopDog (approved):

PR #1301, opened 2026-06-01
feat: persistent user data layer (favorites, saved analyses, workflow run tracking)

dannon:
## What this does

Adds a DB-backed user persistence layer on top of the already-merged Keycloak SSO foundation:

- **Favorites** -- users can save assembly entities; "Saved" state survives reload and is scoped per account.
- **Saved analyses** -- assistant chat sessions can be snapshotted, listed at `/assistant/saved`, and restored back into a fresh working session.
- **Workflow run tracking** -- every Galaxy handoff (from site browse or from the assistant) records a row in `workflow_runs`, so users see their launch history at `/account/workflow-runs` and the `launch_source="assistant"` analytics correlate back to the originating chat session.
- **Account menu + preferences page** -- a header dropdown gates the four new pages; the preferences page is a placeholder that persists a raw JSON object, confirming the account-settings path works end to end.

## Architecture

New Postgres tables under `backend/api/app/db/`: `users`, `favorites`, `saved_analyses`, `workflow_runs`. SQLAlchemy 2.x async + Alembic migrations. The Keycloak OIDC callback upserts the user row, with a guard against the concurrent-first-login race (two simultaneous first requests no longer collide on the insert -- the loser catches `IntegrityError`, rolls back, refetches, and updates).

New API endpoints under `/api/v1/`:
- `auth/me`, `user/me`, `user/preferences`
- `favorites` (GET/POST/DELETE)
- `saved_analyses` (GET/POST/DELETE/restore)
- `workflow_runs` (GET/POST)

Frontend: account menu component, four new pages (`/data/favorites`, `/assistant/saved`, `/account/workflow-runs`, `/account/preferences`), an `AssemblyFavoriteButton` (disabled while a toggle is in flight, so a fast double-click can't fire racing add/remove calls off a stale snapshot), a `useAssemblyFavorites` hook, and a `brc-api-client` service for talking to the new endpoints. `useAssistantChat` and `useLaunchGalaxy` are extended to thread saved-analysis IDs and workflow-run tracking through.

Server-side ownership is enforced from the auth session, never from request bodies. Tests cover cross-user isolation (`test_user_persistence_api.py`, `test_workflow_runs_api.py`) and the assistant-session ownership check on the workflow-runs POST.

Anonymous-to-authenticated handoff: a chat session started while logged out can be claimed by the user when they sign in. `SessionService.claim_session()` stamps the user's sub onto an anonymous session, gated on possession of the signed session cookie -- so in deployed environments (where `SESSION_COOKIE_SECRET` is set; the gate no-ops in local/dev when it's unset) knowing the session ID alone is not enough to claim it. Applied to `/assistant/chat` (continue) and `/saved_analyses` POST (save), and covered by a session-cookie-gate test on the save path (valid cookie -> 200, missing/wrong -> 403).

## Testing performed

- Full local gate is green on the current tip: Prettier, ESLint, `tsc --noEmit`, backend `uv run pytest` (220 passing; the 8 `test_links.py` smoke tests need a live Docker backend and are excluded), and the frontend Jest suite (555 passing). Branch is rebased onto current `main`.
- Drove the four flows end to end in a browser against the `docker-compose.auth.yml` stack (Keycloak + Postgres + Redis + Galaxy): SSO login + user upsert, favorites add/remove/persist, workflow-runs API + listing. The saved-analyses save path was exercised via API + DB (the assistant UI save flow needs an LLM key the test stack doesn't carry).
- Verified server-side ownership: POSTing `/favorites` with a fabricated `user_id` body field is ignored; the favorite lands on the authenticated user.
- Verified the workflow_runs ownership fix in the live stack: a bogus `assistant_session_id` is rejected with 404.

## Notes for reviewers

- DB migrations are off by default on startup. The `docker-compose.auth.yml` stack sets `RUN_MIGRATIONS_ON_STARTUP=true` for local/dev/CI; prod deploys keep explicit migration control via `alembic upgrade head` at deploy time.
- `workflow_runs.user_id` uses `ON DELETE SET NULL` on purpose -- the table mixes authenticated and anonymous tracking, so a deleted user's run gracefully becomes an anonymous record rather than being lost.
- `compute_handoff()` carries the assistant session ID into Galaxy handoff URLs as `?assistantSessionId=...`. That's what wires the `launch_source="assistant"` correlation together; the frontend `useLaunchGalaxy` reads it back when recording the run.
- This introduces an app database: `DATABASE_URL` is now required for the persistent-user features (`get_engine()` raises if it's unset), and `docker-compose.auth.yml` adds a `brc-app-db` Postgres service alongside the migration knob.
- The full husky gate (Prettier / ESLint / `tsc` + `ruff format`) now passes on the branch tip, which is rebased onto current `main` on top of the merged Keycloak auth.

## Out of scope for this PR

- `/account/preferences` is a self-acknowledged placeholder until product picks settings to wire up.
- The API `version` field still reports the Dockerfile default rather than `package.json` -- separate hygiene fix.
- Anonymous POSTs to `/workflow_runs` aren't rate-limited yet (audit flagged it; rate-limiting is its own change). The body-size cap added here handles the payload-abuse angle; request-frequency is separate.
- True pre-parse request-size protection (middleware / proxy limits). The Content-Length checks in `/user/preferences` and `/workflow_runs` are authoritative for storage size, but FastAPI has already parsed the body before they run -- deferred to an ingress-layer change.
- A handful of small audit nits (entity_type query validation, mid-word title truncation, saveAnalysis error-message granularity) deferred to keep this PR focused.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -5,7 +5,15 @@
 coverage/*
 
 venv
+**/venv/*
+.venv
+**/.venv/*
+.pytest_cache
+.ruff_cache
 coverage
+
+# GitHub Actions CommonJS helpers (not project source)
+.github/actions/**/*.js
 
 # generated
 /catalog/schema/generated
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,4 +1,5 @@
 {
+  "root": true,
   "parser": "@typescript-eslint/parser",
   "ignorePatterns": [
     "next-env.d.ts",
--- a/.prettierignore
+++ b/.prettierignore
@@ -13,6 +13,8 @@ coverage
 # python
 venv
 .venv
+.pytest_cache
+.ruff_cache
 
 # releases
 /CHANGELOG.md
--- a/app/components/Assistant/ChatPanel/chatPanel.tsx
+++ b/app/components/Assistant/ChatPanel/chatPanel.tsx
@@ -7,6 +7,7 @@ import {
   Typography,
 } from "@mui/material";
 import { JSX, useEffect, useRef, useState } from "react";
+import { useAuth } from "../../../providers/authentication";
 import { SuggestionChip } from "../../../types/api";
 import { ChatMessage } from "../ChatMessage/chatMessage";
 import { SuggestionChips } from "../SuggestionChips/suggestionChips";
@@ -23,19 +24,46 @@ interface ChatPanelProps {
   loading: boolean;
   messages: ChatMessageDisplay[];
   onRetry?: () => Promise<void>;
+  onSave: () => void;
   onSend: (message: string) => void;
+  saveLabel: string | null;
+  saveLoading: boolean;
   suggestions: SuggestionChip[];
 }
 
+/**
+ * The main chat interface with message list, input, and suggestion chips.
+ * @param props - Component props
+ * @param props.error - Error message to display
+ * @param props.isRestoring - Whether a previous session is being restored
+ * @param props.loading - Whether the assistant is processing
+ * @param props.messages - Chat message history
+ * @param props.onRetry - Callback to retry the last failed request
+ * @param props.onSave - Callback to save the current chat
+ * @param props.onSend - Callback to send a message
+ * @param props.saveLabel - Save status message
+ * @param props.saveLoading - Whether a save request is in flight
+ * @param props.suggestions - Suggestion chips to display
+ * @returns Chat panel element
+ */
 export const ChatPanel = ({
   error,
   isRestoring,
   loading,
   messages,
   onRetry,
+  onSave,
   onSend,
+  saveLabel,
+  saveLoading,
   suggestions,
 }: ChatPanelProps): JSX.Element => {
+  const {
+    isAuthenticated,
+    isConfigured,
+    isLoading: isAuthLoading,
+    login,
+  } = useAuth();
   const [input, setInput] = useState("");
   const messagesEndRef = useRef<HTMLDivElement>(null);
 
@@ -65,6 +93,21 @@ export const ChatPanel = ({
   };
 
   const inputDisabled = loading || !!isRestoring;
+
+  const handleSave = (): void => {
+    if (isConfigured && !isAuthLoading && !isAuthenticated) {
+      login();
+      return;
+    }
+    onSave();
+  };
+
+  let saveButtonLabel = "Save";
+  if (saveLoading) {
+    saveButtonLabel = "Saving...";
+  } else if (isConfigured && !isAuthenticated) {
+    saveButtonLabel = "Sign In to Save";
+  }
 
   return (
     <ChatContainer>
@@ -127,6 +170,15 @@ export const ChatPanel = ({
       />
 
       <InputRow>
+        <Button
+          disabled={
+            loading || saveLoading || messages.length === 0 || isAuthLoading
+          }
+          onClick={handleSave}
+          variant="outlined"
+        >
+          {saveButtonLabel}
+        </Button>
         <TextField
           disabled={inputDisabled}
           fullWidth
@@ -146,6 +198,13 @@ export const ChatPanel = ({
           Send
         </Button>
       </InputRow>
+      {saveLabel && (
+        <Box sx={{ pb: 2, px: 2 }}>
+          <Typography color="text.secondary" variant="body2">
+            {saveLabel}
+          </Typography>
+        </Box>
+      )}
     </ChatContainer>
   );
 };
--- a/app/components/Entity/components/ConfigureWorkflowInputs/components/Main/components/Stepper/components/Step/hooks/UseLaunchGalaxy/useLaunchGalaxy.ts
+++ b/app/components/Entity/components/ConfigureWorkflowInputs/components/Main/components/Stepper/components/Step/hooks/UseLaunchGalaxy/useLaunchGalaxy.ts
@@ -1,5 +1,7 @@
 import { useAsync } from "@databiosphere/findable-ui/lib/hooks/useAsync";
 import { useConfig } from "@databiosphere/findable-ui/lib/hooks/useConfig";
+import { brcAPIClient } from "app/services/brc-api-client";
+import { useRouter } from "next/router";
 import { useCallback } from "react";
 import { Workflow } from "../../../../../../../../../../../../apis/catalog/brc-analytics-catalog/common/entities";
 import {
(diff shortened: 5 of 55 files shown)


dannon:
Thanks @copilot-pull-request-reviewer -- two real findings here and both were worth acting on. Good signal on this one.

**Real bug, fixed:**
- **Cross-user session access returned 503 instead of 403** ([0210c18d](https://github.com/galaxyproject/brc-analytics/pull/1301/commits/0210c18d88e531de3b9449ce82a1d655fd754a4f)). You flagged this from both sides -- `chat()` wrapping `PermissionError` into a generic `RuntimeError` (`assistant_agent.py`) and the route mapping that to 503 (`assistant.py`). Correct: the authenticated cross-user case already 403s via the claim block, but an *anonymous* caller citing a session owned by someone else skipped that and hit `chat()`, so the authz failure surfaced as a transient-outage 503. `PermissionError` now propagates and maps to 403, with a regression test.

**Cleanup, not a bug -- fixed anyway because it's cheap:**
- **`launch_source` taken verbatim from the request body** ([98d2f7a8](https://github.com/galaxyproject/brc-analytics/pull/1301/commits/98d2f7a8f51b4047193a4bc7ae2dedd8ba0b844d)). True in principle, but the only client posting here is our own frontend, which already derives `"assistant"`/`"site"` from session presence (`useLaunchGalaxy/utils.ts:243`) -- so no real run is mislabeled today, and the security-relevant check (assistant-session ownership) was already enforced. Still worth hardening: the server now derives `launch_source` from the ownership-checked `assistant_session_id` rather than trusting the body, so a hand-crafted request can't claim assistant attribution it doesn't have. Test added.

Side note: each comment arrived twice, and the 503 issue was split across two adjacent lines/files -- counted as two findings regardless.

Both fixes are backend-only; the affected suites are green (23 tests, +2 new). Genuinely useful round -- 8/10.

NoopDog:
Thanks for this, @dannon! What needs to be done to get this out of draft?

dannon:
@NoopDog Basically everything in here needs identity (users), so we need to land the auth setup.  I would like to have that functional, and tested against prior to merging.  The good news is that it is getting a lot of love this week and I expect it in the next day or two.

dannon:
@NoopDog auth's landed now, so this is unblocked. Rebased onto current `main`, full gate's green (backend + frontend tests + husky), and the earlier Copilot findings are already in. Taking it out of draft -- ready for a look.

dannon:
Thanks @copilot-pull-request-reviewer -- both fair this round.

**Fixed:**
- **`persistence_app` leaked the global DB engine** -- client tests reset it via the app lifespan shutdown, but the fixture used directly (the concurrent-insert race test) left `_engine`/`_session_factory` cached and bled into the next test. Added a `close_db()` teardown. ([4d8b71a6](https://github.com/galaxyproject/brc-analytics/commit/4d8b71a6))
- **Misleading Content-Length comment in `update_preferences`** -- right that the body's already parsed by the time the handler runs, so it isn't saving Pydantic any work. Reworded as a best-effort early 413 from the header, with the serialized-size check still authoritative. ([75db937e](https://github.com/galaxyproject/brc-analytics/commit/75db937e))

Two for two -- good round.

dannon:
Thanks @copilot-pull-request-reviewer -- all three landed.

**Fixed:**
- **`/assistant/chat` continued sessions without requiring the cookie** -- only the authenticated claim path checked it, so a leaked `session_id` could be replayed. Now any request carrying a `session_id` requires the signed cookie, matching what `restore`/`delete` already enforce. ([17347e8e](https://github.com/galaxyproject/brc-analytics/commit/17347e8e))
- **Stale preferences status banner** -- the load effect now clears `error`/`saveMessage`, so a transient failure doesn't linger after a later success or a sign-out/in. ([ea8418b4](https://github.com/galaxyproject/brc-analytics/commit/ea8418b4))
- **Favorites empty-state flash** -- `isLoading` now defaults to true (and resets in the unauthenticated path), so an authenticated page shows a spinner instead of briefly flashing "no favorites yet." ([e707e130](https://github.com/galaxyproject/brc-analytics/commit/e707e130))

Three for three this round -- nice catches, especially the session-cookie gap.

dannon:
Thanks @copilot-pull-request-reviewer -- most of this round is against an earlier state of the branch; a lot is already handled in the code, sometimes with a comment saying exactly what you suggest.

**Already handled -- premise doesn't hold against current code:**
- **PermissionError -> 503 not 403** -- `chat()` already lets `PermissionError` propagate (it only catches `KeyError`), and the route maps it to 403 *before* the RuntimeError->503 handler.
- **`launch_source` taken verbatim** -- it's derived server-side; `"assistant"` requires a real ownership-checked session, and a spoofed value with no session is downgraded to `"site"`.
- **Fixture leaks the DB engine** -- there's a `close_db()` teardown disposing the cached engine.
- **`isLoading` starts false / flicker** -- it initializes to `true` and the unauth early-return sets it `false`.
- **Content-Length comment misleading** -- the comment already says FastAPI has parsed the body by then and this is just a fast 413.
- **Cookie only enforced for auth-claiming** -- the cookie is required for *any* `request.session_id`, unconditionally.
- **preferences error not cleared** -- the effect clears `error`/`saveMessage` before fetching.

**Fair but minor -- noting, not blocking:**
- **"gates re-entry" comment** -- correct, `setIsToggling(true)` is a UI flag, not a real guard. Comment fix.
- **ARIA on the menu trigger** -- fair, could add `aria-haspopup`/`aria-expanded`/`aria-controls`.
- **`MAX_WORKFLOW_RUN_BYTES` scope** -- the serialized check bounds `parameters`; the Content-Length guard covers the rest. Comment could be tighter.
- **No Jest test for `buildWorkflowRunPayload`** -- fair, real branching worth covering.
- **`router.query` useCallback dep** -- fair micro-perf nit; a derived stable `assistantSessionId` is cleaner.

Seven of twelve describe problems the code already solves (and several were posted 2-3x).

mvdbeek (approved):
Looks good to me. Claude was a little concerned about the default dev deployment, but it does seem like a standard pattern:

```
  3. Session-cookie binding silently no-ops without SESSION_COOKIE_SECRET

  backend/api/app/core/session_signing.py: the new require_session_cookie returns early (allow-all) when no secret is set,
  and set_session_cookie only marks the cookie Secure outside DEV_ENVIRONMENTS (which defaults to development). A deploy
  that forgets SESSION_COOKIE_SECRET and/or ENVIRONMENT therefore runs with binding fully disabled and/or an insecure
  cookie, with no startup error. The docstring says this is intentional "opt-in" — but for a security control, failing open
  silently is a foot-gun. Fix: require the secret (and a non-dev ENVIRONMENT) to be set explicitly in non-dev, or log a
  loud warning at startup.
```

Issue #1296, opened 2026-05-31
Assistant: stop offering ENA/SRA search it can't do — route to the ENA picker at workflow setup


NoopDog:
Parent epic: #1288

When selecting read runs, the assistant offers "Search for Public ENA SRA data," then responds it can't search ENA/SRA — it has no ENA tool (the prompt frames `data_source` as "user upload or ENA/SRA public data" at `assistant_agent.py:122` and suggests "Search ENA for public data" at `:201`). Don't offer a capability we don't have.

offering to search.. 
<img width="1999" height="1022" alt="Image" src="https://github.com/user-attachments/assets/7f76d50f-0c66-4419-95a1-394bdda59801" />

then declining..
<img width="761" height="507" alt="Image" src="https://github.com/user-attachments/assets/faca39e2-84f1-49b1-a49e-15a868eba75f" />

Workflow setup already provides the ENA picker: the Sequencing step supports `VIEW.ENA` (enter an accession / select from ENA) and `VIEW.UPLOAD_MY_DATA` (`SequencingStep.tsx:44-77`); the handoff passes `data_source` through (`SchemaPanel` → handoff; `WorkflowInputsView.tsx:50`).

**Change:** remove the assistant's offer to search ENA/SRA. For public data, let the user **Continue to Workflow Setup** and use the existing ENA picker there. Fallback: have the assistant ask for ENA accessions directly instead of offering to search.

## Acceptance criteria

- The assistant no longer offers a search capability it doesn't have.
- A user wanting public ENA data reaches it via the ENA picker at workflow setup (or by providing accessions directly to the assistant).

d-callan:
need to coordinate a bit. dannon has something here going already

dannon:
The only thing still making it *offer* to search is the prompt framing of `data_source` as "user upload or ENA/SRA public data" (`assistant_agent.py:122`). I'll tweak that so it stops offering and points people to the ENA picker at workflow setup instead -- small prompt PR, I'll grab it.

Issue #723, opened 2025-08-19
ENA picker refinements


nekrut:
# Case 1: Finding SRA data for a particular organism

In this scenario a user selects an organism and proceeds to configure Galaxy inputs clicking "Browse all 466 Sequences"

> [!CAUTION]
> "Browse all 466 Sequences" should **really** say  "Browse all 466 Sequences for this organism"

<img width="852" height="340" alt="Image" src="https://github.com/user-attachments/assets/928b72ea-e977-410b-ace0-f1b13d613ed1" />

This given this table:

<img width="1693" height="1179" alt="Image" src="https://github.com/user-attachments/assets/3e55c262-29aa-48aa-bbde-4f7b2fe1fdc4" />

One of the first things one needs here is to figure out what BioProjects are available

<img width="333" height="425" alt="Image" src="https://github.com/user-attachments/assets/d60a409b-911f-4ec8-ad41-f4be8b356c36" />

Here are important things to make it very useful:

- Naturally this list needs to be sorted by size in the descending order
- Project IDs should be clickable and these links should go to something like this https://www.ncbi.nlm.nih.gov/bioproject/?term=PRJNA1174053, which will open in a new Tab of the browser. This will allow quickly decide whether this is relevant or not
- Can columns of the Run viewer be sortable? (even better: can some summary stats [max,min] be shown there?)
- It should be possible to download the table as a tsv,csv
- We need to add additional validation for paired-end data: In some cases the datasets sas it is paired but it actually not:

<img width="1441" height="120" alt="Image" src="https://github.com/user-attachments/assets/c4f28b74-44fc-4cfa-aeaa-cbab66719b8d" />

Here ERR2352345 is paired but SRR175719 is not. We see that by the number of entries in "Fastq FTP" field. In other words, when we select "Paired" in the filter we need to make sure that "Library Layout" == "Paired" and len("Fastq FTP".split(',')) > 1. Otherwise we get this:

<img width="683" height="422" alt="Image" src="https://github.com/user-attachments/assets/0e887e3a-0e1d-464d-9642-58f513053b4d" />

NoopDog:
Thanks!

nekrut:
> Interesting, the PAIRED is wrong or only has a single direction.

It is unfortunately very common problem. I don't know what is wrong (mislabeling or forgot to upload) but it not right

PR #1302, opened 2026-06-01
feat: add SRA metadata connector (DuckDB mirror), a faster interim before SOLR

dannon:
## What this does

Adds a new sequencing-metadata **connector** alongside the existing ENA connector: a read-only SRA run-metadata mirror backed by a local DuckDB file. It's wired into both the in-app assistant and the MCP server, so users (and external MCP clients) can ask catalog-wide questions -- "how much data exists for *M. tuberculosis*?", "largest BioProjects for *P. falciparum*?", "recent Nanopore runs since last year?" -- grounded in real numbers, sub-second, with no external API on the request path.

This is positioned as a **low-commitment interim before the planned SOLR search backend**. It's strictly opt-in: with `SRA_MIRROR_PATH` unset (the default), `is_available()` returns False, no tools register, no prompt mentions them, and every existing deploy behaves exactly as before. When SOLR lands, these same tools can swap their backend without touching the agent or MCP surface.

## Why -- improving the connectors

Today's sequencing search leans entirely on the live ENA connector (`ENAService`), which queries the EBI portal API per lookup. That's fine for fresh single-accession reads but weak for the discovery/aggregate queries BRC users actually want: ENA is rate-sensitive on broad queries (results capped at 300), has no aggregate endpoints, needs Python-side post-filtering for multi-faceted queries, and an EBI outage takes SRA search down with it.

NCBI publishes the full SRA metadata table as daily-refreshed public Parquet on AWS Open Data. We pre-filter that to BRC catalog taxids (expanded through the NCBI taxonomy tree, and -- critically -- resolved through `taxdump/names.dmp` so synonyms are recovered in both directions) into a compact local DuckDB file. The connector is a thin read-only wrapper over it: summary stats, filtered run search, top-BioProject rollups, and study/BioProject-accession lookups, all in well under a second.

It sits **side by side with `ENAService`, not as a replacement** -- ENA stays the right call for fresh-data accession lookups; the mirror is the right call for aggregates, filters, and cohort discovery. Both are interim until SOLR.

## Addresses #1047 (snapshot provenance / ID drift)

@d-callan's #1047 is the right worry: external IDs drift, and if you don't pin a snapshot you don't know what you matched. This connector answers that by construction -- every tool response carries a `_meta` block with the mirror build date, the taxdump version, and the actual SRA name strings the query resolved to (after taxid expansion). So a reply reads like "based on the SRA snapshot of 2026-05-23, queried under both *Candida auris* and *Candidozyma auris*..." The provenance contract from #1047 is baked in, not bolted on later.

(That synonym resolution isn't cosmetic -- naive name-matching silently drops huge amounts of data in both directions. Resolving every catalog taxid against `names.dmp` at build time recovered ~7.76M runs across the catalog that a naive match would have missed.)

## What's in this PR

- `app/services/sra_mirror.py` (new) -- `SRAMirrorService`, a read-only DuckDB wrapper with taxid-anchored name resolution, a curated abbreviation alias map (SARS-CoV-2, TB, HIV, etc. that NCBI doesn't name), an in-process bounded TTL cache, and the `_meta` provenance block on every response.
- `app/services/tools/sra_tools.py` (new) -- 4 pydantic-ai tools for the assistant: `sra_summary_for_organism`, `search_sra_runs`, `top_bioprojects_for_organism`, `get_sra_study_runs`. Registered on the agent only when the mirror is available.
- `app/services/mcp_server.py` -- 3 of those exposed on `/api/v1/mcp` (`search_sra`, `sra_data_summary`, `get_sra_study_runs`), gated the same way, with SRA-vs-ENA routing guidance added to the server `instructions`. (`top_bioprojects` is intentionally not exposed over MCP -- the summary already embeds the top 10.)
- Opt-in wiring in `core/config.py`, `core/dependencies.py`, and `main.py`, plus the matching guard in `evals/tasks.py`.
- Evals: two datasets (`sra_tool_selection`, `sra_assistant_multiturn`) including negative cases that lock in "the new SRA tools don't poach catalog-only queries," with a dataset guard that surfaces "requires SRA_MIRROR_PATH" instead of silently scoring 0.0.
- Docs: a new "## MCP Server" section in `backend/api/README.md` covering catalog, ENA, and the opt-in SRA tools.

## Hardening done before this PR

A self-review + multi-angle bug review produced a punch list, all addressed and individually TDD'd:

- **Opt-in correctness** -- the system prompt no longer advertises tools that aren't registered; `dependencies.py` gates construction on a non-empty path; the `Path('').exists()` trap (which logged a scary traceback on every default cold start) is closed.
- **Query correctness** -- deterministic taxid resolution (`ORDER BY taxid`); a `resolved` flag so a typo'd organism reads differently from a real organism with no data; `since`-date validation with a polite error instead of a crashed turn; case-insensitive country matching with a synonym map; normalized accession casing for study/BioProject lookups.
- **Robustness/hygiene** -- `_initialize` splits IO/missing-table/error cases and closes the handle on failure (no leaked file lock); the cache is bounded (512-entry cap + eviction), deep-copies on store/read, and keys on a normalized organism string; service-layer limit clamps as defense in depth; stable sort tiebreakers so rows at the LIMIT boundary don't shuffle between calls.

## Testing

- Backend suite green at 205 tests at last run (the only excluded failures are the pre-existing `test_links.py` network tests that need a live server). New coverage in `tests/test_sra_mirror.py`, `tests/test_dependencies.py`, `tests/test_mcp.py` (13 new), and `evals/tests/test_sra_dataset_guard.py`.
- `uv lock --check` passes (duckdb pinned at 1.5.3).
- Committed in a worktree without root `node_modules`, so the husky frontend gate hasn't run -- this PR is backend-only so that's a formality, but flagging it.

## Out of scope (deliberately)

- **The mirror artifact and its ingest pipeline.** This PR is the read-only consumer. The `.duckdb` file (taxdump-resolved filter of the SRA parquet) is built externally and not checked in; the connector does nothing until `SRA_MIRROR_PATH` points at one.
- **Deployment/refresh story** -- where the file lives on the deployed backend, the weekly ingest cron, taxdump rebuild cadence. Tracked separately.
- **Connection-refresh on atomic swap** -- the service holds one long-lived read-only connection, so an external mirror swap stays invisible until restart. Deferred with the freshness work.
- **Replacing `ENAService`, and SOLR itself** -- the mirror coexists with ENA and is explicitly the interim until SOLR ships.

## Related issues

- Addresses #1047 (snapshot provenance / external ID drift) -- see the provenance section above.
- Answers the catalog-shape aggregate queries from #723.
- Relates to #1296 ("stop offering ENA/SRA search the assistant can't do"). The SRA tools and their prompt guidance only register when the mirror is available, so the assistant stops advertising SRA search it can't perform -- and when the mirror is set, the capability is genuinely there. The ENA-offer wording and the route-to-the-picker fallback that #1296 also calls for are separate and not touched here.

### DIFF
--- a/backend/api/README.md
+++ b/backend/api/README.md
@@ -72,6 +72,23 @@ Services:
 - `GET /api/docs` - Interactive Swagger UI
 - `GET /api/redoc` - ReDoc API documentation
 
+## MCP Server
+
+The API embeds a Model Context Protocol server at `/api/v1/mcp`, giving AI
+clients (Claude Desktop, the Galaxy MCP integration) direct access to the BRC
+catalog and sequencing-data search.
+
+**Catalog tools** -- organisms, assemblies, workflows, and compatibility checks
+(in-memory, always available).
+
+**ENA tools** (`search_ena`, `search_ena_keywords`) -- live sequencing-run
+search against the European Nucleotide Archive.
+
+**SRA mirror tools** (`search_sra`, `sra_data_summary`, `get_sra_study_runs`) --
+fast structured search over a local SRA metadata mirror scoped to BRC-relevant
+organisms. Opt-in: registered only when `SRA_MIRROR_PATH` points at a built
+mirror file; a default deploy exposes only the catalog and ENA tools.
+
 ## Configuration
 
 Environment variables (see `.env.example`):
--- a/backend/api/app/core/config.py
+++ b/backend/api/app/core/config.py
@@ -122,6 +122,9 @@ class Settings:
         # Catalog path
         self.CATALOG_PATH: str = os.getenv("CATALOG_PATH", "/catalog/output")
 
+        # SRA-DuckDB mirror. Empty path disables the assistant's SRA tools.
+        self.SRA_MIRROR_PATH: str = os.getenv("SRA_MIRROR_PATH", "")
+
         # Keycloak / OIDC settings
         self.KEYCLOAK_ISSUER_URL: str = os.getenv(
             "KEYCLOAK_ISSUER_URL",
--- a/backend/api/app/core/dependencies.py
+++ b/backend/api/app/core/dependencies.py
@@ -1,5 +1,6 @@
 import logging
 from functools import lru_cache
+from typing import Optional
 
 from fastapi import Cookie, Depends, HTTPException, Request
 from sqlalchemy.ext.asyncio import AsyncSession
@@ -49,8 +50,11 @@ def get_ena_service() -> ENAService:
 
 
 @lru_cache(maxsize=1)
-def get_sra_mirror_service() -> SRAMirrorService:
+def get_sra_mirror_service() -> Optional[SRAMirrorService]:
     settings = get_settings()
+    if not settings.SRA_MIRROR_PATH:
+        logger.info("SRA_MIRROR_PATH not set -- SRA mirror service disabled")
+        return None
     service = SRAMirrorService(settings.SRA_MIRROR_PATH)
     logger.info(
         f"SRA mirror service initialized (singleton), available: {service.is_available()}"
--- a/backend/api/app/core/dependencies.py
+++ b/backend/api/app/core/dependencies.py
@@ -15,6 +15,7 @@ from app.services.assistant_agent import AssistantAgent
 from app.services.auth_service import COOKIE_NAME, AuthService
 from app.services.catalog_data import CatalogData
 from app.services.ena_service import ENAService
+from app.services.sra_mirror import SRAMirrorService
 
 logger = logging.getLogger(__name__)
 
@@ -48,9 +49,20 @@ def get_ena_service() -> ENAService:
 
 
 @lru_cache(maxsize=1)
+def get_sra_mirror_service() -> SRAMirrorService:
+    settings = get_settings()
+    service = SRAMirrorService(settings.SRA_MIRROR_PATH)
+    logger.info(
+        f"SRA mirror service initialized (singleton), available: {service.is_available()}"
+    )
+    return service
+
+
+@lru_cache(maxsize=1)
 def get_assistant_agent() -> AssistantAgent:
     cache = get_cache_service()
-    agent = AssistantAgent(cache)
+    sra_mirror = get_sra_mirror_service()
+    agent = AssistantAgent(cache, sra_mirror=sra_mirror)
     logger.info(
         f"Assistant agent initialized (singleton), available: {agent.is_available()}"
     )
@@ -130,5 +142,6 @@ def reset_all_services() -> None:
     get_cache_service.cache_clear()
     get_catalog_data.cache_clear()
     get_ena_service.cache_clear()
+    get_sra_mirror_service.cache_clear()
     get_assistant_agent.cache_clear()
     get_rate_limiter.cache_clear()
--- a/backend/api/app/main.py
+++ b/backend/api/app/main.py
@@ -24,6 +24,7 @@ from app.core.dependencies import (
     get_cache_service,
     get_catalog_data,
     get_ena_service,
+    get_sra_mirror_service,
     reset_all_services,
 )
 from app.db.session import close_db, init_db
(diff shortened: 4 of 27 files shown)


dannon:
Thanks @copilot-pull-request-reviewer -- three findings, all worth fixing and no noise this time. Nice round.

**Behavior fix:**
- **`platform`/`assay_type` were case-sensitive** ([d3d61336](https://github.com/galaxyproject/brc-analytics/pull/1302/commits/d3d61336fec8b32945dde508aa15a2fc49f32089)). Good catch -- `country` got case-insensitive matching in an earlier hardening pass but these two didn't, so `platform="illumina"` silently returned zero rows against the stored `"ILLUMINA"`. Both now match case-insensitively, normalized the *same* way for the SQL filter and the search cache key -- which also covers the cache-key half of your comment, and avoids a subtle window where a whitespace-padded value could cache a zero-result under the shared key. Tests cover the lowercase match, the padded-input case, and assay-type exclusion.

**LLM-facing docstring fix:**
- **`country` docstring said "exact country name"** ([0a44942e](https://github.com/galaxyproject/brc-analytics/pull/1302/commits/0a44942eb61fbd52576bf9e7b7fde7153eb78f8a)). Stale after the synonym/case work -- it was nudging the model to over-normalize the input. Updated to say it's matched case-insensitively with common synonyms.

**Eval-harness fix:**
- **`primary_score` computed against `_CASES`, not the filtered cases** ([c9bd3f91](https://github.com/galaxyproject/brc-analytics/pull/1302/commits/c9bd3f91bfee960f2490612a84348cf49d04c1a7)). Right -- a `--only` run selecting cases with no `expected_schema` still resolved to `FinalSchemaContains`, which those cases never attach, so the primary metric wouldn't surface. Now computed from the cases actually built. (Checked the sibling `sra_tool_selection` dataset -- it doesn't derive primary_score this way, so nothing to change there.)

All backend-only; the SRA mirror + eval-guard suites are green (38 tests, +2 new). Solid round -- 9/10.

NoopDog:
Hi @dannon, very interesting. Thanks.  What needs to be done to get this out of draft?  Also looks like it needs a rebase. Cheers, D

dannon:
Thanks @copilot-pull-request-reviewer -- mixed bag, and a chunk of it is reviewing an earlier state of the branch.

**Fixed:**
- **Dev-local docstring path** -- the `sra_mirror` module docstring pointed at a path in my personal notes; replaced it with a high-level description of the external build. ([73bf6736](https://github.com/galaxyproject/brc-analytics/commit/73bf67363a5c93dbdbe46c6a4bdd2014cf629ce7))

**Fair but minor -- noting, not blocking:**
- **`since` cache key** -- true, the key uses the raw `since` while the query uses the normalized value, so `"2021"` and `"2021-01-01"` land in separate cache entries. Duplicate entries, never a wrong result.
- **Invalid-`since` error shape** -- that early return omits the `_meta`/`resolved` block the success path carries. Minor contract inconsistency.

**Already handled on the branch (older diff):**
- **Country docstring "exact name"** -- already says matched case-insensitively, synonyms accepted (UK/USA).
- **`primary_score` vs `_CASES`** -- already computed against the built `cases` list.
- **Filter normalization / case-sensitive platform+assay SQL** -- filters are normalized in the cache key and the SQL is `LOWER(...)` on both.

**Not applicable:**
- **JSON vs plain-string return** -- the consumer is the LLM tool-call layer; the unavailable-path prose is intentional.
- **`exit 1` on all-skipped datasets** -- nothing in CI invokes `run_evals`; exit-nonzero-on-no-output is fine for a manual runner.

Three of eight were already fixed in the branch.

dannon:
Thanks @copilot-pull-request-reviewer -- two this round, one fair:

- **`since` error message** -- good catch. The text only listed `YYYY-MM-DD`, but `_normalize_since` also accepts `YYYY` and `YYYY-MM`, so the error was steering the model away from formats that actually work. Fixed the wording. ([8d5c465c](https://github.com/galaxyproject/brc-analytics/commit/8d5c465c2d61a1f434ced65178c649d2e1e194a5))
- **`_norm_filter` empty vs whitespace** -- technically true that `""` and `"   "` end up handled a hair differently, but nothing realistically passes a whitespace-only filter through the tool, so leaving it.

mvdbeek (approved):

PR #1307, opened 2026-06-02
feat: add assembly count min and max workflow annotations

### LINKED ISSUE #1306: Add assembly count constraints (min/max) to workflow model for organism-scoped workflows
so some organism scoped workflows require no assemblies (bc they build an assembly for ex) and some require many (comparative genomics stuffs).. in order to tell the difference and build out steppers accordingly, or maybe conditionally show workflows fi they need many assemblies and an organism only has one, etc.. we need to annotate them.

d-callan:
## Description

Adds `assembly_count_min` and `assembly_count_max` constraints to the workflow model and uses them to filter organism-scoped workflows by available assembly count.

## Related Issue
 
closes https://github.com/galaxyproject/brc-analytics/issues/1306

### DIFF
--- a/app/apis/catalog/brc-analytics-catalog/common/entities.ts
+++ b/app/apis/catalog/brc-analytics-catalog/common/entities.ts
@@ -117,6 +117,8 @@ export interface WorkflowCategory {
 }
 
 export interface Workflow {
+  assemblyCountMax: number | null;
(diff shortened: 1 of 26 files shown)


hunterckx (approved):
Thanks @d-callan!

Issue #1299, opened 2026-05-31
Assistant: dedicated Anthropic API keys for dev and prod with sufficient limits and budget refill


NoopDog:
Parent epic: #1288

Light single-user testing in a single session hit a hard usage cap:

```
BadRequestError: Error code: 400 - {'type': 'error', 'error': {'type': 'invalid_request_error',
'message': 'You have reached your specified API usage limits. You will regain access on
2026-06-01 at 00:00 UTC.'}, 'request_id': 'req_011CbaBTjUt4uNEUadUUoSEN'}
```

This is a **usage/budget cap** on the Anthropic key (`AI_API_KEY`), not a per-minute rate limit — it blocks access until the reset date. Light usage shouldn't reach it, which suggests the key may be shared with another service.

**Do:**
- Provision a dedicated Anthropic API key for **dev** and a separate one for **prod** (not shared with other services).
- Set sufficient usage limits / budget with automatic refill for each.
- Configure each environment to use its own `AI_API_KEY`.

## Acceptance criteria

- Dev and prod each use a dedicated, separate API key.
- Limits/budget are sufficient for beta usage and refill automatically; light testing no longer hits the cap.

NoopDog:
Do we have keys now that are independent for dev and prod and not shared with anyone? I noticed the dev assistant is working again, and we have a mention in the PR above. But it's not clear if this can be closed yet?

PR #1308, opened 2026-06-02
feat: ground the assistant's suggestion chips in the catalog

### LINKED ISSUE #1297: Assistant: only offer selections for data in the catalog (don't suggest organisms we don't have)
Parent epic: #1288

Image

The assistant offers options for data that isn't in the catalog. Example: it asks "Which Candida species are your samples from? (e.g., C. albicans, C. auris, C. glabrata)" and presents a selectable chip for *C. glabrata*; selecting it then returns that *C. glabrata* is not in the catalog. We should not offer selections for data we don't have.

The suggestion-chips prompt (`assistant_agent.py:196-204`) asks for "phrases the user might want to say next" with no requirement that they be grounded in the catalog; offered options also appear inline in the model's prose.

Determine an approach to only offer data the catalog actually has (organisms, assemblies, workflows), review it, then implement.

## Acceptance criteria

- An agreed approach for grounding offered selections in catalog data.
- The assistant no longer offers selections for organisms/data not in the catalog.

dannon:
Closes #1297.

The assistant was offering specific organisms it didn't have -- e.g. "which Candida species?" with a clickable C. glabrata chip that then turned out not to be in the catalog. Two layers:

**1. Prompt rule.** Forbids presenting a specific organism/assembly/workflow -- in prose or as a chip -- unless a tool returned it this conversation; tells the model to search first instead of listing examples from memory. The Candida case is called out explicitly.

**2. Server-side chip validation (defense-in-depth).** Suggestion chips can tag the entity they propose (`{"label": "...", "assembly": "GCF_000002765.6"}`); the parser validates tagged organism/assembly/workflow refs against the catalog and drops any that don't resolve. Organisms are matched *exactly* (not via the fuzzy search), keys are case-insensitive, entity values are coerced before lookup (so a numeric taxid is validated, not skipped), and empty tags are treated as untagged. Plain action chips pass through; the frontend still consumes label/message, so no frontend change.

This is defense-in-depth, not a hard guarantee: the tag is validated but the visible label is free text, so ungrounded mentions in prose or untagged chips stay prompt-controlled. That matches the issue's "offer selections" framing.

Reviewed adversarially by Codex over two rounds -- it caught fuzzy-match false positives, key-normalization bypasses, and a regression where non-string tags skipped validation; all tightened with matching test coverage. An eval for the prompt behavior is a follow-up: it needs the eval harness to expose per-turn suggestions plus an actual model run, which is gated on #1299's API budget.

### DIFF
--- a/backend/api/app/services/assistant_agent.py
+++ b/backend/api/app/services/assistant_agent.py
@@ -216,8 +216,9 @@ prefixed with "SUGGESTIONS:". Each item is either:
 
 Only tag a chip with an entity a tool actually returned -- never offer a \
 specific organism, assembly, or workflow you haven't verified. Tagged chips \
-that don't match the catalog are dropped before the user sees them, so an \
-invented entity simply won't appear. Generic action chips don't need a tag.
+whose entity isn't in the catalog are dropped before the user sees them, and \
+the label must name the same entity you tagged. Generic action chips don't \
+need a tag.
 
 Example:
 SUGGESTIONS: ["Tell me about variant calling", {"label": "Use the \
@@ -658,6 +659,8 @@ class AssistantAgent:
                 if not self._chip_entities_in_catalog(item):
                     logger.info("Dropping ungrounded suggestion chip: %s", label)
                     continue
+                # The tag is validated above; the label itself is free text the
+                # prompt is responsible for keeping consistent with the tag.
                 chips.append(SuggestionChip(label=label, message=label))
             # Anything else (numbers, nested lists) is ignored.
         return chips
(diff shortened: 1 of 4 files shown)


dannon:
Thanks @copilot-pull-request-reviewer -- 3 for 3, all real. Fixed in 2b119548:

- **Null chip label:** a `{"label": null, ...}` chip was being coerced to the literal "None"; now only real string labels are accepted, anything else is dropped.
- **Trailing-space keys:** entity tag keys are now stripped before lowercasing, so `"organism "` no longer slips past validation as an untagged chip.
- **`find_organism_exact` signature:** widened to `Any` to match the numeric-taxid / `None` inputs it already coerces and handles.

Added tests for the null-label and trailing-space-key cases.

backend/api/app/services/assistant_agent.py
```diff
@@ -617,6 +636,80 @@ def _parse_structured_output(
...
+        contract -- a chip that tags a display name instead will be dropped.
+
+        This is defense-in-depth, not a hard guarantee: it validates the tag, but
+        the visible label is free text and only constrained by the prompt. Any
+        lookup error fails closed (drop the chip).
+        """
+        # Strip + lowercase keys so "Organism" or "organism " don't bypass.
+        tags = {k.strip().lower(): v for k, v in chip.items() if isinstance(k, str)}
+        lookups = (
+            ("organism", self.catalog.find_organism_exact),
```
d-callan, line 696: if feasible, a part of me would prefer to just depend on the taxon ids.

d-callan (approved):

backend/api/app/services/assistant_agent.py
dannon: Good call -- done in 8e3bd2ad. The chip contract now tags organisms by NCBI taxonomy id instead of the species name (the model already has the taxid from tool output, so it's basically free), and the validator keys on the taxid. I kept exact species/common-name matching as a fail-soft fallback so we don't needlessly drop a legitimately grounded chip if the model tags a name anyway -- but the canonical path is the taxid now. Shout if you'd rather I rip the name fallback out and go taxid-only.

NoopDog (approved):

PR #1338, opened 2026-06-09
feat: about page reorg

### LINKED ISSUE #560: update about page
i think id actually ideally like multiple pages, or at least table of contents type thing to left, idk. but one way or another it feels like some important info needs added and made easier to find/ navigate. what i personally see as desired content:
1. vision statement, though we dont need to use those words on the site. just something that says 'why'. right now the page has a lot of 'what' and 'who' and just a bit of 'how'.. if that makes sense
2. 'who' is currently under the title 'collaborative strategy' and only focuses on PIs. feels like we could add others, even if only list them. (see [ex](https://www.bv-brc.org/team))
3. there should be a collaborative strategy section, probably calling for _community collaboration_ and instructions/ links to how to contribute iwc workflows for ex, or where to go if you want to request an organism be added, etc. 
4. the blurbs under each of the four 'what are these resources'  bits are too long. no one cares. not really anyway, sry. they want to know what role each plays in this project maybe, but not your whole history. im not even convinced we need this section tbh, at least not as it is. a diagram or something like is below makes more sense to me personally.
5. there should be a bit about how we fit into other BRCs. i think @NoopDog had a ticket for that somewhere we should maybe reference here
6. the 'collaborative strategy' section also has some 'how' type info, starting to lean in to technical stuffs like managing references via ucsc. that info should be clarified, given its own section, maybe better organized, fleshed out more etc.
7. it should be very obvious, probably at top or even in multiple places on the page, **HOW TO CONTACT US**. we have links to github, to discourse and to a google survey thingy. none of these links have words, or make it clear when you should follow one route or another etc. (aside: who is watching these things or gets notifications when users engage them?)
8. maybe an acknowledgements section? particularly given the collaborative community strategy, it probably makes sense to acknowledge ppl and orgs who arent directly funded/ involved but contributed something meaningful

happy to discuss, have others leave comments w things they think make sense on the about page, break into sub-issues as needed etc.
also just a note, that once we have published some things, we should have clear citation instructions and listed publications

d-callan:
## Description

roadmap under about, old about page becomes 'partner resources' for both brc and ga2. for brc also gains a page under about w high level who we are and what are we doing, that i even dared call a vision lol. small section at bottom of landing that now feels redundant removed. a few spots where existing content was modified slightly for things like shifting from future to present tense.

## Related Issue

closes #560


frano-m (approved):
LGTM! Thanks @d-callan 🧇

Issue #1192, opened 2026-03-12
Epic: Introduce a Database Backing the Catalog Pipeline


d-callan:
so here is a thing i think we should at least be talking about soonish

## Problem

The app currently relies entirely on large JSON artifacts built offline (catalog/output, catalog/ga2/output). As the data grows and relationships get richer, JSON-only storage and simple Python scripts are creating fragility:
- External IDs can change upstream; JSON snapshots get stale without lineage tracking.
- Relationships (workflows ↔ assemblies ↔ organisms ↔ read runs) are loosely defined across multiple files.
- Read-run counts follow different generation rules than the rest of the catalog.
- Some data is shared between BRC and GA2, some is site-specific.
- Build scripts are long-running and monolithic; no modular pipeline to rebuild small slices.
- Filtering logic (workflow compatibility) is duplicated: in the build, in workflow page, and in assembly detail pages.
- API “filters” loop through JSON arrays, which doesn’t scale as entity counts grow.

## Why Now

Upcoming features will stress the current pipeline:
- Ingesting real SRA metadata that needs search/filter + relationships to existing assemblies.
- New “genes” entity with many-to-many relationships to assemblies/organisms (orders of magnitude more rows).
- Different “support levels” for assemblies (FASTA only vs. indexes) requiring contextual display rules.
- Large influx of assemblies → table/filter performance concerns.
- User accounts with saved entities or custom submissions (“my assemblies”, “add organism by taxon ID”) will need stable identifiers and referential integrity.
- The JSON build will only get longer and harder to reason about; moving to a DB/API unlocks incremental updates, validations, and eventual automation.

## Goals

1. Design a persistent catalog store (e.g., Postgres) that models workflows, assemblies, organisms, read runs, genes, etc., with clear foreign keys.
2. Define ETL pipeline to populate/refresh the DB from upstream sources with modular steps and reproducible logging.
3. Expose data through an API consumed by the Next.js app instead of reading JSON bundles.
4. Consolidate filtering/compatibility logic so it runs once in the DB/ETL layer; clients get precomputed relationships.
5. Support site-specific catalogs (BRC vs. GA2) via filters/views in the DB, not separate JSON directories.
6. Lay groundwork for user-associated data (saved assemblies, custom organisms) stored alongside catalog objects.

NoopDog:
Filed #1366 as a **temporary stopgap** for one symptom this epic will ultimately fix.

[...]

PR #1355, opened 2026-06-15
fix: update some taxon ids for upstream changes at ncbi

### LINKED ISSUE #1350: Align species-specific workflow taxonomy_ids with catalog organism taxids (influenza 11320→2955291; audit sars-cov-2/pox)
Sub-issue of #1200 (Epic: Support Organism-Scoped Workflows with Collection Inputs).

### Problem

Species-specific workflows in `catalog/source/workflows.yml` pin `taxonomy_id` to **legacy NCBI nodes that are not organism entries** in the catalog, while the catalog keys organisms on the new (2024 ICTV-aligned) species taxids. Concretely, the influenza workflow uses `11320` ("Influenza A virus"), but in the catalog that node is now nested *below* the species **`2955291` ("Alphainfluenzavirus influenzae")**, which is the actual organism entry (7 assemblies).

Catalog lineage (from an Influenza A assembly, `GCF_000851145.1`):

```
… → 197911 Alphainfluenzavirus (genus)
      → 2955291 Alphainfluenzavirus influenzae (species)  ← organisms.json key
          → 11320 Influenza A virus (legacy node)         ← workflows.yml taxonomy_id
              → 102796 → 130760 (isolate)
```

So `11320` is a **descendant** of `2955291`, not an ancestor — NCBI inserted the formal species above the legacy node.

### Why it matters

- Workflow↔assembly **compatibility still works** for all of these via the lineage check (#1328 "wf_tax in lineage"), so nothing is visibly broken today.
- But **#1349** needs to resolve a workflow's `taxonomy_id` *to an organism entry* (RHS organism panel + ENA read-run picker). `11320` does not resolve to an organism; `2955291` does. Aligning the workflow taxid to the catalog organism key makes that a direct hit and is more future-proof (catches isolates tagged at the species but not under the legacy node).

### Fix

1. Change the influenza workflow `taxonomy_id` **`11320` → `2955291`** in `catalog/source/workflows.yml` (line ~1015), then rebuild so `catalog/output/workflows.json` updates. **(organisms list already uses `2955291` — correct, no change.)**
2. **Audit the other species-specific workflows** for the same legacy/ICTV-rename mismatch and align to the catalog organism taxid:
   - `sars-cov-2-*` workflows → `taxonomy_id: 694009` (not an organism entry; likely should be the new SARS-CoV-2 species `3418604` — verify against `organisms.json`)
   - `pox-virus-amplicon-main` → `taxonomy_id: 10244` (verify)
3. Leave the **broad-clade gates** alone — `2` (Bacteria), `2759` (Eukaryota), `10239` (Viruses), `131567`, `11158` are intentionally high-rank scopes, not single organisms.

### Out of scope

`catalog/source/outbreaks.yml` also uses `11320`, but deliberately — its NCBI/BV-BRC virus-browser URLs key on `taxid:11320`. Don't change it here.

### Acceptance criteria

- [ ] Influenza workflow `taxonomy_id` is `2955291` in source and rebuilt output; it resolves to the "Alphainfluenzavirus influenzae" organism entry.
- [ ] sars-cov-2 and pox workflow taxids audited and aligned to their catalog organism taxids (or confirmed correct with a note).
- [ ] Compatibility for the affected workflows verified unchanged after the swap (lineage match still returns the same assemblies).

d-callan:
## Description

so flu was the one pointed out, but while i was there i thought: qc report should have caught this, lets make sure it does next time. so there are also some new qc report entries in various places. and that helped me realize mpox had a similar issue to flu, so fixed that as well.

of note: hep c changed viral family apparently. it used to be in priority pathogens and now isnt. almost happened without my noticing, which is a bit disconcerting. but i think the change is in this case fortunately actually fine and i dont think i have a good way of catching that in the qc report the way the build process works currently. something to reconsider when we get to a db that can hopefully track data changes over time maybe: #1192 

## Related Issue

closes #1350


hunterckx (approved):
Thanks @d-callan!

PR #1356, opened 2026-06-15
feat: user chip account menu + fix account page layouts

dannon:
Refreshes the account UI now that login is live on dev, and picks up two things at once.

The signed-in header used to be a plain name button with the user's email floating next to it as raw text -- it looked out of place and crowded the nav on narrow widths. This swaps it for an avatar chip that opens the account menu, with name and email in a header. The menu keeps the links that landed while this branch sat (Favorites, Saved Analyses, Workflow Runs, Preferences) plus Sign out.

Separately, the four account pages were rendering hero and content side by side instead of stacked -- they never set a Main layout, so they fell through to findable-ui's default row Main. Setting StyledPagesMain like every other content page fixes the stacking and the page background.

Rebased onto current main (the branch was ~120 commits behind, from before the Next 16 / findable-ui v54 bump).

### DIFF
--- a/app/components/Layout/components/Header/components/AuthButton/authButton.styles.ts
+++ b/app/components/Layout/components/Header/components/AuthButton/authButton.styles.ts
@@ -1,8 +1,39 @@
-import styled from "@emotion/styled";
+import { Button, styled } from "@mui/material";
 
-export const AuthButtonWrapper = styled.div`
-  align-items: center;
-  display: flex;
-  gap: 8px;
-  white-space: nowrap;
-`;
+// Clickable, pill-shaped user chip (avatar + name + caret) that anchors the
+// account menu. Reads as a grouped control rather than loose text next to a
+// button, and visually separates the account from the nav links.
+export const UserChip = styled(Button)(({ theme }) => ({
+  // The avatar is the Button's startIcon, whose default rule
+  // (.MuiButton-startIcon > :nth-of-type(1)) forces a 20px font on its child and
+  // ties a plain .MuiAvatar-root override on specificity. Scope to the startIcon
+  // context so the initials actually shrink to fit the 28px circle.
+  "& .MuiButton-startIcon .MuiAvatar-root": {
+    backgroundColor: theme.palette.primary.main,
+    color: theme.palette.primary.contrastText,
+    fontSize: 12,
+    fontWeight: 600,
+    height: 28,
+    width: 28,
+  },
+  "& .MuiButton-startIcon, & .MuiButton-endIcon": {
+    margin: 0,
+  },
+  "&:hover": {
+    backgroundColor: theme.palette.action.selected,
+  },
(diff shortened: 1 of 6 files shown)


dannon:
Thanks @copilot-pull-request-reviewer -- all three were fair, and I applied them.

**Accessibility -- the menu trigger** ([25200128](https://github.com/dannon/brc-analytics/commit/25200128caffb40551416325b1a22217a06e3d31)):
- Gave the chip an `id`, switched `aria-haspopup` from the generic boolean to `"menu"`, and now only set `aria-expanded` while the menu is open.
- Wired the menu's list to the trigger via `aria-labelledby` so screen readers announce it in the context of the chip.

**Theme tokens** ([b5a1804a](https://github.com/dannon/brc-analytics/commit/b5a1804ad2f9f2e683b3a928faa428bc41505d59)):
- Replaced the hard-coded `rgba(0, 0, 0, ...)` background/hover with `theme.palette.action.hover` / `action.selected`. Same values in the current theme, but they now track it -- matches how the rest of the app styles its hover states.

Good signal this round -- nothing to push back on.

dannon:
Thanks @copilot-pull-request-reviewer -- two real catches, two with an off premise. All four addressed in 4531af46.

**Valid menu markup -- fixed (good catch):** MUI `Menu` renders its children inside a `<ul>`, so the header `<div>` and the `<Divider>`'s `<hr>` were invalid list children. The header now renders as an `<li>` (`UserMenuHeader`) and the divider via `<Divider component="li" />`. Verified the rendered menu's `<ul>` now contains only `<li>` children (header, separator, menuitem).

**The `&` scoping -- applied, but the premise is off:** emotion already scopes bare nested selectors to the component class -- the emitted rule was `.css-…-MuiButton-root .MuiAvatar-root`, not a global `.MuiButton-startIcon .MuiAvatar-root`, so nothing was leaking onto other buttons. The explicit `&` is clearer regardless, so I added it to the startIcon/endIcon selectors.

dannon:
Thanks @copilot-pull-request-reviewer -- both of these were fair, applied them.

**`useId()` for the menu IDs** -- the collision premise (two `AuthButton`s on one page) doesn't actually fire today, since it's a single header control. But deriving the button/menu IDs from `useId()` is the right idiom and costs nothing, so done. ([6dd3386](https://github.com/dannon/brc-analytics/commit/6dd33867))

**Menu header as `role="presentation"`** -- good catch. A bare `<li>` sitting in MUI's `<ul role="menu">` isn't a valid menu child; marking the non-interactive header `role="presentation"` keeps it out of the menu's ARIA hierarchy without touching the markup. ([202566e](https://github.com/dannon/brc-analytics/commit/202566ed))

Four rounds on a two-file PR is a lot of bites at the same apple, but these last two landed. **Copilot review: 7/10 this round.**

d-callan (approved):
i see conflicts but they look simple

PR #1377, opened 2026-06-25
refactor: catalog build: count ena read runs via /count instead of downloading every run (#1366)

### LINKED ISSUE #1366: Read-run eligibility: check the count live via ENA /count at the stepper
## Problem
The read-run eligibility gate is driven by a **build-time precompute**. `generate_taxon_read_run_count` (`catalog/py_package/catalog_build/build.py:1932`) calls the ENA portal **`search`** endpoint, downloads **every** `read_run` row for `tax_tree(taxid)`, and takes `len()` — across ~2,000 taxids (TB subtree ≈ 349k runs) that's **millions of rows pulled just to count**. The counts ship as `taxonomy_read_counts.json`, which the runtime gate `isEligible()` (`app/components/Entity/.../Stepper/components/Step/SequencingStep/components/ENASequencingData/query/utils.ts`) reads to decide whether to offer "browse all" ENA data (`hook.ts`: `enabled = isEligible(ncbiTaxonomyId, maxReadRunsForBrowseAll)`).

Result: the count is a **between-builds snapshot** (drifts from live ENA) and costs a slow scan of the whole catalog to produce — all to gate one UI decision.

## Approach: check the count live on the stepper
Switch the gate to a **live** lookup. When the user reaches the sequencing step for a selected taxon, fetch the count from ENA's dedicated `/count` endpoint (through the existing proxy already used for the data fetch, `NEXT_PUBLIC_ENA_PROXY_DOMAIN`):

```
{ENA_PROXY}/ena/portal/api/count?result=read_run&query=tax_tree({taxonomyId})
```

Parse the integer and apply the same `maxReadRunsForBrowseAll` threshold to drive `isEligible`. One cheap live call per stepper visit, always fresh.

This is a **frontend-only** change: `isEligible` consults the live count instead of the bundled JSON. The build-time precompute is left in place (now unused) — see deferral below.

## Why it works
- `/count` returns in **~0.6s flat regardless of magnitude** (server-side aggregate; measured TB ~311k and *H. sapiens* both ~0.64s) — fine for an on-demand call on the step.
- Only the **one selected taxon** is counted per visit, not ~2,000 — trivial work.
- `tax_tree(...)` keeps counts subtree-inclusive — same semantics as today.
- Always live — no staleness; the frontend already hits ENA live for the actual rows (`queryFn.ts`), so the gate stops being the one stale, precomputed piece.

## Considerations
- Adds a **runtime dependency on ENA** for the gate: keep it **non-blocking** (resolve the count alongside, don't gate page render) and degrade gracefully on timeout/failure (defined fallback — e.g. default the eligibility decision). Carry over the existing `ConnectTimeout`/retry posture to the client call.
- ~0.6s added latency on the step; optionally cache per-taxon for the session.
- Decide the fallback when the count is unavailable (allow vs. withhold "browse all").

## Scope / non-goals
- Switches the **eligibility gate** to a live `/count` call; that's it.
- **Removing the precompute is a follow-on ticket** — out of scope here. `generate_taxon_read_run_count`, the BRC + GA2 `taxIdReadCount.json` steps, `catalog/build/py/mergeBRCandGA2ReadCounts.py`, and the generated `taxonomy_read_counts.json` are **left untouched**; they simply go unused once the gate is live and get deleted in the follow-on. (Keeps this change small and frontend-only.)
- Does **not** change the threshold semantics (`maxReadRunsForBrowseAll`) or the downstream ENA data fetch.
- Does **not** touch taxonomy resolution / taxdump / organism resolution (separate tickets).

## Acceptance criteria
- [ ] The sequencing step obtains the read-run count via a **live ENA `/count`** call (through the ENA proxy) for the selected taxon.
- [ ] `isEligible` uses the live count and no longer reads `taxonomy_read_counts.json`.
- [ ] The count call is **non-blocking** and degrades gracefully on ENA error/timeout (defined fallback).
- [ ] Spot-check eligibility decisions for a few taxa (small, large, borderline).
- [ ] Build pipeline untouched (precompute removal tracked in the follow-on).

## Follow-on (separate ticket)
Once the gate is live, delete the now-unused precompute and its outputs: `generate_taxon_read_run_count`, both `taxIdReadCount.json` build steps, `mergeBRCandGA2ReadCounts.py`, and `taxonomy_read_counts.json`.

## Context — relationship to #1192
End state is **#1192 (Database Backing the Catalog Pipeline)**: once ENA/SRA read-run data is in a queryable DB, the count comes from the DB (with real pagination). A live ENA `/count` on the stepper is the natural stopgap already shaped like that end state (query on demand). (Live ENA `search` has no `offset`, so true paged *listing* still needs the DB — but *counting* doesn't, which is why the gate can move live now.)

frano-m:
## Summary

Closes #1366.

`generate_taxon_read_run_count` computed per-taxon read-run counts by hitting ENA's **`/search`** endpoint with `result=read_run` and **downloading every row** just to `len()` them — millions of rows across ~2,000 catalog taxids (e.g. *M. tuberculosis* subtree ≈ 349k), a major drag on the Phase-1 build (`npm run build-brc-from-ncbi`).

This switches the count to ENA's dedicated **`/count`** endpoint, which returns the integer as a server-side aggregate.

## Change

`catalog/py_package/catalog_build/build.py` — `generate_taxon_read_run_count`:
- `url`: `/ena/portal/api/search` → `/ena/portal/api/count`
- dropped the now-unneeded `fields` param
- `taxon_counter[tId] = len(resp.json())` → `int(resp.json()["count"])` (`/count?format=json` returns `{"count": "<n>"}`)
- `ConnectTimeout` retry handling preserved

Shared helper, so **both BRC and GA2** builds benefit (both call `create_taxonomy_read_run_count` → `generate_taxon_read_run_count`).

## Why it's safe & fast

- `/count` returns in ~0.6s regardless of magnitude (server-side aggregate, not a fetch).
- `tax_tree(...)` query shape is unchanged → counts stay subtree-inclusive and equivalent.
- `/count?result=read_run` counts the same result set `len(search(result=read_run))` did → equivalent values (modulo live drift).

## Verification

Ran the real function against live ENA:
- `2955291 → 67,705`, `10090 → 4,022,917`, `562 → 595,874` — all ints, sorted desc.
- `ruff format` clean; no new `ruff check` errors introduced.

## Scope / non-goals

- Only the read-count **generation** step; `taxIdReadCount.json` / `taxonomy_read_counts.json` schema unchanged (taxid → int).
- Runtime gate `isEligible()` untouched (still reads precomputed JSON).
- Temporary stopgap until #1192 (DB-backed catalog pipeline) makes this precompute step disappear entirely.

### DIFF
--- a/catalog/py_package/catalog_build/build.py
+++ b/catalog/py_package/catalog_build/build.py
@@ -1931,7 +1931,10 @@ def create_taxonomy_read_run_count(genomes_tsv_path: str, output_path: str):
 
 def generate_taxon_read_run_count(taxonomy_ids):
     taxon_counter = {}
-    url = "https://www.ebi.ac.uk/ena/portal/api/search"
+    # Use ENA's dedicated /count endpoint, which returns the read-run count as a
+    # server-side aggregate (~0.6s regardless of magnitude) instead of the
+    # /search endpoint, which would download every read_run row just to len() it.
+    url = "https://www.ebi.ac.uk/ena/portal/api/count"
     counter = 0
     num_taxids = len(taxonomy_ids)
     for tId in taxonomy_ids:
@@ -1942,7 +1945,6 @@ def generate_taxon_read_run_count(taxonomy_ids):
         params = {
             "result": "read_run",
             "query": f"tax_tree({tId})",
-            "fields": "experiment_accession,study_accession",
             "format": "json",
         }
         try:
@@ -1953,7 +1955,8 @@ def generate_taxon_read_run_count(taxonomy_ids):
             time.sleep(10)
             resp = requests.get(url, params=params)
             resp.raise_for_status()
-        taxon_counter[tId] = len(resp.json())
+        # /count?format=json returns {"count": "<n>"}.
+        taxon_counter[tId] = int(resp.json()["count"])
         counter += 1
     print(f"Processed {counter} taxonomy IDs", end="\n")
     return dict(sorted(taxon_counter.items(), key=lambda x: x[1], reverse=True))


frano-m:
Closing — #1366's design has changed (per team discussion). Rather than optimizing the catalog-build precompute (`search`→`count`), the new direction is to query ENA `/count` **live in the frontend** when the stepper loads (replacing the `taxonomy_read_counts.json` lookup), and remove the build-time precompute as a separate follow-on. That makes this build-side optimization moot, since `generate_taxon_read_run_count` is slated for removal. Re-scoped #1366 will land as the frontend live-count change.

PR #1399, opened 2026-07-02
feat: add new world screwworm assemblies

d-callan:
## Description

adding some assemblies for parasitic flies currently seen in outbreak impacting texas


d-callan:
two things:
1. i had to update taxon ids again for some changed to a few virus
2. the qc report says the new ones arent available at ucsc, but ive seen them there so.. i think a configuration thing, which ill check w max and hiram about. its ok if this sits until i hear back

d-callan:
think this is ready now @hunterckx :pray: . while i was here getting in sync w ucsc, i un-commented a few assemblies that are available there now from a prev request. also there were a few hiram said he cant get to build properly that i removed.

hunterckx (approved):
Thanks @d-callan!

Issue #1473, opened 2026-07-21
Prerender real content for entity detail pages (organism/assembly) instead of empty client-hydrated shells


NoopDog:
## Problem

The entity **detail** pages (`/data/assemblies/{id}`, `/data/organisms/{id}`) are prerendered as **empty shells**. Their built `<body>` has 0 visible characters — all content is rendered client-side after hydration. So the ~7,500 detail pages we build get no SEO and no server first-paint; a crawler that doesn't run JS sees a blank page, and users wait for client JS + a data fetch before anything appears.

Notably, the data is already available at build time: `getStaticProps` for the detail route calls `processEntityProps` → `seedDatabase` and sets `props.data` with the entity record. The build has everything it needs to render — the architecture just doesn't use it for output.

## Why it renders empty (two coupled blockers)

1. **The gate is tied to a client-only effect.** `pages/_app.tsx` wraps pages in `EntitiesLoadedProvider`, whose boolean comes from `useEntities()`:
   ```ts
   const [isLoaded, setIsLoaded] = useState(false);   // starts false
   useEffect(() => { ensureEntitiesLoaded(config).then(() => setIsLoaded(true)); }, [config]);
   ```
   Effects never run during SSG, so `isLoaded` is always `false` at build and `EntityDataGate` renders `null`. This is deliberate — it avoids a hydration mismatch, because the data arrives via a client-only lifecycle.

2. **The views read from a `fetch()`-populated client store, not from `props`.** `ensureEntitiesLoaded` → `loadWorkflows`/`loadEntities` populate `app/services/workflows/store` via `fetch("/api/*.json")`. The detail views (`EntityDetailView`, `AnalyzeView`) read synchronously from *that* store (`getAssembly()`, `getOrganism()`, `getWorkflows()`), which is empty at build. They ignore the `props.data` that `getStaticProps` already provides.

There are effectively two parallel data systems: the findable-ui `database` singleton + `props.data` (populated at build) and the BRC `app/services/workflows/store` (populated by client fetch). The detail views are wired to the second.

## Proposed work

Make the entity **detail** views prerender real content:

1. **Source detail views from build-available data** — `props.data` (already passed) and/or a build-time-populated store — instead of the fetch-only client store.
2. **Render those routes server-side** (drop/relax `EntityDataGate` for detail pages).
[...]

PR #1483, opened 2026-07-21
refactor: back the priority-pathogens explore page with the client entity store (#1479)

### LINKED ISSUE #1479: Back the priority-pathogens explore page with the client entity store
## Background

In #1472 the assemblies and organisms explore list pages stopped inlining the catalog into their static HTML and instead source their rows from the client-loaded entity store (`getEntitiesByType`, populated by `loadEntities` / `useEntities()`).

Priority pathogens was split onto its own static page (`data/priority-pathogens`) but, unlike assemblies/organisms, it was **kept on page-props** (`seedDatabase` + `fetchAllEntities`) because it is not part of the client entity store. Its data comes from `staticLoadFile: catalog/output/outbreaks.json`, which is not one of the routes the loader fetches. As a result, its static HTML still inlines its data (~184 KB).

## Proposal

Add priority pathogens (outbreaks) to the loader/store so its explore page can source rows from the client store like assemblies and organisms:

- Add a route/API entry for outbreaks (mirror `app/services/workflows/routes.ts`), packaging `outbreaks.json` as a runtime `/api/*.json`.
- Wire it into `loadEntities` so `getEntitiesByType` returns it.
- Repoint `data/priority-pathogens` to render via the store (as `EntitiesView` does), dropping its page-props `seedDatabase` / `fetchAllEntities`.

Note: `Outbreak.description` is a pre-compiled `MDXRemoteSerializeResult` (server-serialized), so confirm it round-trips cleanly through the client store / explore table.

## Follow-up from

#1472 (stop inlining the catalog into the assemblies/organisms explore pages).

frano-m:
## Summary

The priority-pathogens explore page was the last data explore page still inlining its full catalog — the outbreak data, including pre-compiled MDX descriptions, was serialized into the static HTML (~184 KB). It now sources its rows from the client-loaded entity store, matching the assemblies/organisms explore pages:

- **`app/services/workflows/routes.ts`** — add `"priority-pathogens": "/api/priority-pathogens.json"` so `loadEntities` fetches + seeds it (it already has a `getId`).
- **`scripts/sync-api-brc-analytics.sh`** — copy `catalog/output/outbreaks.json` → `public/api/priority-pathogens.json` (the catalog output name differs from the route). BRC-only; GA2's sync is untouched.
- **`PriorityPathogensView`** — reads the store (`[...getEntities<Outbreak>("priority-pathogens")].sort(...)`, shallow-copied so the sort can't mutate the shared store); dropped its `data` prop; removed the now-empty `types.ts`.
- **pp explore page** — gated on `EntityDataGate` (the store read is deferred past static prerender); `getStaticProps` reduced to `{ entityListType, meta }` + the `notFound` guard for sites without priority-pathogens (GA2).

## Result

| | Before | After |
|---|---|---|
| `data/priority-pathogens.html` | ~184 KB | ~37 KB |

Detail pages (24) unchanged. `Outbreak.description` (`MDXRemoteSerializeResult`) round-trips through the client fetch and renders correctly (verified in-browser: cards, priority chips, and MDX body all render).

## Verification

tsc / eslint / prettier / jest (581) / build all green; no inlined outbreak data remaining in the explore HTML; GA2 build unaffected.

Closes #1479

image

### DIFF
--- a/app/services/workflows/routes.ts
+++ b/app/services/workflows/routes.ts
@@ -2,6 +2,7 @@ export const API = {
   assemblies: "/api/assemblies.json",
   organisms: "/api/organisms.json",
   pangenomes: "/api/pangenomes.json",
+  "priority-pathogens": "/api/priority-pathogens.json",
   workflowAssemblyMappings: "/api/workflow-assembly-mappings.json",
   workflows: "/api/workflows.json",
 } as const;
--- a/app/views/PriorityPathogensView/priorityPathogensView.tsx
+++ b/app/views/PriorityPathogensView/priorityPathogensView.tsx
@@ -1,19 +1,19 @@
+import type { Outbreak } from "@/apis/catalog/brc-analytics-catalog/common/entities";
+import { getEntities } from "@/services/workflows/query";
 import { useLayoutSpacing } from "@databiosphere/findable-ui/lib/hooks/UseLayoutSpacing/hook";
 import { JSX } from "react";
 import { PriorityPathogens } from "./components/PriorityPathogens/priorityPathogens";
 import { StyledGrid, StyledTitle } from "./priorityPathogensView.styles";
-import { Props } from "./types";
 import { sortPriorityPathogen } from "./utils";
 
-export const PriorityPathogensView = (props: Props): JSX.Element => {
+const ENTITY_LIST_TYPE = "priority-pathogens";
+
+export const PriorityPathogensView = (): JSX.Element => {
   const { spacing } = useLayoutSpacing();
 
-  // Get priority pathogens data.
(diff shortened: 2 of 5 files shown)


frano-m:
Closing — priority pathogens is a small (24-item) static card list, so keeping it fully pre-rendered (content in the HTML → SEO + first paint) is better than a client-store shell. The store-backing here was only for consistency with the assemblies/organisms lists, but those are the special case (~15 MB catalogs + interactive filterable tables) where client rendering pays off. `main` already renders pp statically (from #1478), so no change is needed. Aligns with #1473 (prerender real content instead of empty shells).

PR #1520, opened 2026-07-25
feat: use all common names provided by ncbi for each assembly taxon (#1431)

### LINKED ISSUE #1431: Refine common name handling
As highlighted by #1396, the common name that we save for an assembly taxon is somewhat arbitrary, as NCBI may provide multiple common names and we only use the first (with type `genbank common name` included and prioritized as of #1396). This may make searching by common name less reliable. We should likely either save all common names, or, if we find some value in continuing to only save one, determine a method of ensuring that the names we choose stay consistent between builds.

The current common name narrowing happens in `get_species_df` in `build.py`.

## Prerequisites

Should be based on #1396 (ticket #1367) -- EDIT: which is now merged

hunterckx:
## Description

Updates singular common name fields to instead hold multiple common names, derived from the name lists provided by NCBI

Notes:
- The common name table columns in the app (which are hidden by default and mainly there for filtering) are updated to display the lists of names in N-tag cells
- Common names for an organism are aggregated from its assemblies
- Unlike existing list fields that are saved in the intermediate TSV (which are just comma-separated), common names are saved as JSON, since as far as we're concerned they may contain arbitrary characters
- Both catalogs have been rebuilt, which for GA2 is just to get the TSV up-to-date; it doesn't include common names in the ultimate JSON output
  - NCBI has moved what is now *Megascapheus bottae* to its new genus, so I've updated the taxonomy ID in the GA2 source to avoid dropping assemblies from the output
- "null" is replaced with "None" as the common name value added to a workflow when an assembly lacks a common name; this was a Claude edit that seemed reasonable enough to me

## Related Issue

Closes #1431


NoopDog (approved):

d-callan (approved):

frano-m (approved):
FE looks good to me thanks @hunterckx

Issue #1294, opened 2026-05-31
Assistant: verify and enable logging of user queries and assistant responses


NoopDog:
Parent epic: #1288

A goal of the beta is to watch what prompts users enter and how the assistant responds. We need that captured somewhere reviewable before go-live.

**Current state (from the source):** the assistant logs per-turn metadata only — run start/elapsed (`assistant_agent.py:450,457`) and token usage (`assistant_agent.py:523`); `api/v1/assistant.py` logs only exception paths. The user query and assistant reply text are **not** logged — they live only in the ephemeral Redis session state (`state.messages`).

**Design step:** determine what to log (e.g. query, reply, session id, timestamp), where to log it, how the logs will be made available for review, and the retention/privacy approach — how long it's kept, who can access it, and whether any user notice is needed.

**Then:** implement the agreed capture.

## Acceptance criteria

- We can review real user queries and assistant responses during the beta period.
- The capture mechanism, storage location, retention, and privacy approach are documented and reviewed as acceptable for production.

NoopDog:
Design context for the logging approach (from a backend investigation):

**Current state**

- Full conversation content (`result.all_messages()`, including tool calls) is persisted only to Redis via `SessionService`, key `assistant:session:{id}`, `SESSION_TTL = 7200` (2h), flushed on startup. No durable record — conversations are unrecoverable after ~2h.
- App logs use stdlib `logging` to stdout only; only token counts are logged, not content.
- Sentry SDK (`sentry-sdk[fastapi]`, `traces_sample_rate=1.0`) active when `SENTRY_DSN` is set — captures errors and request transactions, not LLM content.
- Agent (`pydantic-ai==1.68.0`) constructed without instrumentation (no `instrument=True`, OpenTelemetry, or Logfire).
- A `pydantic-evals` harness exists under `evals/`.

**Options**

1. **Flat durable log → Postgres.** At the existing `to_jsonable_python(result.all_messages())` point in `assistant_agent.chat()`, also write a row (session id, timestamp, user message, reply, full message JSON, token counts, latency). Review/charts via Grafana's Postgres datasource. ~20 lines + a table; no new always-on services. No per-step span tree.
[...]

Issue #1293, opened 2026-05-30
Assistant: define key usage metrics and add them to the key metrics report


NoopDog:
Parent epic: #1288

As part of the beta launch, define the key metrics for assistant usage and add them to the key metrics report so it's reviewed alongside the other platform metrics.

As a starting point, consider the assistant funnel: **entry** (starting a session), **exit to the stepper** (handoff), and **launch in Galaxy** — and the conversions between them. Confirm which events are already captured and instrument any that aren't.

## Acceptance criteria

- An agreed set of assistant key metrics.
- The events needed for them are captured.
- Assistant usage appears in the key metrics report.

d-callan:
i think plausible will let us capture when someone visits the assistant, but i agree probably first thing to track is how often that actually results in someone having a configured stepper/ the handoff. we also have info on when someone lands in galaxy from brc, and im happy enough to do things like watch proportion of visits to assembly vs assistant page and upticks in landing in gxy as heuristics for when we land from the assistant for now. if things look off, we can figure how to add proper metrics for that later. 

so high level for a min, to me the things id like to learn from the assistant are:
- handoff from assistant to existing stepper
- when user wants to configure something in right panel and it doesnt work out. ex: an organism or workflow is requested that we dont have yet , no veupath gtf available for that organism
- any time a feature from the site is requested that is outside the assistant scope, like custom analysis and help browsing ucsc tracks
- any time a user has to ask the assistant to undo or remove a configuration from right side panel

and whats interesting to me is that bc this is a chat, extracting some of this info is maybe best done by an llm summarizing the chats we log?

Issue #1288, opened 2026-05-30
[Epic] BRC Assistant: prototype → production (beta launch & beyond)


NoopDog:
The BRC Assistant (in-app chat: explore catalog → configure a workflow → hand off to the stepper/Galaxy) has been running behind a feature flag in prototype mode. This epic tracks turning it on for all production users behind a **`beta` tag**, then iterating from real usage. The `beta` tag sets expectations and lets us ship sooner by deferring non-blocking work; observability lands first so we can see the prompts users enter and the responses we return.

Sub-issues are created and linked in GitHub under the phases below.

## Assistant evaluation rubric

A shared, mostly-objective way to score what the assistant does. Two parts: **what** we test (operation × entity), and **how** each query is scored (graded dimensions + the beta bar).

### What we evaluate

Behavior is an **operation × entity matrix**, not a single ladder. Operations order by complexity; entities (**organism / assembly / workflow**) vary in how well each operation is supported.

Per-entity operations (ascending complexity):

1. **Exact lookup** — the user names a thing precisely (exact name or id) and wants its record. *"Tell me about assembly GCF_000005845.2"*
2. **Resolution** — the user names a thing *imprecisely* (nickname, abbreviation, misspelling, outdated name); the assistant must map it to the right catalog entry before it can look anything up. *"Do you have Candida auris?"* — the catalog holds it under its reclassified name, Candidozyma auris.
3. **Filtered list** — the user wants the subset of records matching one or more attributes. *"Which complete-genome assemblies do you have for Cryptococcus neoformans?"*
4. **Subtree / clade** — the user asks about a whole branch of the taxonomy (a genus, family, …); the answer spans every species beneath that node, not one. *"What assemblies do you have for Anopheles?"* → 69 assemblies across 28 species.
5. **Quantitative / aggregate** — the user wants a computed number or grouping over many records (count, max, group-by), not a list. *"How many chromosome-level assemblies are there?"* → 610.

Relational / global (cross-entity):

6. **Compatibility** — can this workflow actually run on this genome? Matches a workflow's constraints (taxonomy, ploidy) against a chosen assembly/organism. *"What workflows can I run on E. coli?"*
7. **Cross-store** — questions that join the genome catalog to a *separate* data store, the SRA/ENA sequencing-reads database (e.g. how much sequencing data exists for an organism). *"How much RNA-seq data exists for Candida auris?"*
[...]

PR #1527, opened 2026-07-29
feat: remove the assistant feature flag and add a Feedback button

### LINKED ISSUE #1290: Assistant: add a clearly-labeled "Feedback" button on the assistant page
Parent epic: #1288

We're launching the assistant as a beta and want to actively collect user feedback. The site already has a general floating help/support icon that opens our feedback Google Form (`StyledViewSupport`, configured in `site-config/brc-analytics/local/floating/floating.ts`). That icon is generic and easy to miss as a feedback channel.

Add a **clearly-labeled "Feedback" affordance specific to the assistant page** (`/assistant`) so beta users have an obvious way to tell us what's working and what isn't.

- Link it to the **currently configured support/feedback URL** — the same destination the site help icon uses (in `floating.ts`); don't introduce a new URL.
- Make it explicitly read as **"Feedback"** (label, not just an icon), distinct from the generic help icon, and visually associated with the assistant.
- **Placement is a design decision** — to be determined with the designer.
- **Rename the linked Google Form's title to "Help and Feedback"** so it reads correctly when reached from both the general help icon and this new assistant feedback button.

## Acceptance criteria

- A "Feedback" control is visible on the assistant page and clearly identifiable as feedback (not just the generic floating help icon).
- It opens the currently configured feedback form (same as the site help icon), in a new tab.
- The Google Form title reads "Help and Feedback."
- Placement fits the assistant layout and doesn't overlap the chat input or get pushed off-screen (coordinate with #1289).

dannon:
Drops the `assistant` feature flag so the Analysis Assistant is live for everyone as a beta — Phase 1 of #1288 — and gives beta users an obvious way to say how it's going. Closes #1290.

**Merge last.** #1525 fixes a live 503 and #1526 stops deploys wiping sessions; both are worth having in before this turns the feature on for everyone. #1526 also touches `assistantView.tsx` and will conflict textually with this on the `showReset` line — trivial to resolve.

## Removing the flag

The gate comes out entirely rather than getting defaulted to on: `useFeatureFlag` reads localStorage client-side, so a default-on gate would still flash a 404 on first render before hydration. Three gates went — the nav item filter in `_app.tsx`, the 404 in `AssistantView`, and the Learn card filter — along with the flag registration. No `useFeatureFlag("assistant")` consumers remain.

## Feedback button

A labeled "Feedback" button in an actions row under the headline, opposite "New Conversation" — visually part of the assistant, and it can't collide with the chat input or scroll away mid-conversation. It points at the same support form the floating help icon already uses rather than a new destination, opens in a new tab, and says so in its accessible name. That URL is now a `SUPPORT_URL` constant both the floating config and the button import, so the two entry points can't drift.

## Keeping it off GA2

This is the part worth reviewing hardest. Both sites build from one `pages/` tree and the static export ships every page, so **the feature flag was the only thing keeping the assistant off Genome Ark**. Removing it would have put a BRC-branded assistant, BRC metadata and the BRC feedback form on GA2.

Three routes needed guarding, using the `allowedPaths` idiom `/calendar` and the `/about` pages already use:

- `/assistant` — direct guard.
- `/assistant/saved` — its own page, easy to miss. GA2 served the Saved Analyses UI, breadcrumbed to a 404 and stuck on a spinner because auth isn't configured there.
- `/learn/assistant` — `pages/learn/[...slug]` generates a page per file under `app/docs/learn`, so hiding the Learn card wasn't enough; the doc stayed directly reachable. Dropped from the generated paths, with the same check in `getStaticProps` behind it.

Verified on a real GA2 dev server: all three 404 there while `/data/organisms` still renders; on BRC all three still render.

## Copy

Adds a third-party-model disclosure to the chat disclaimer and a "What happens to your messages" section in `assistant.mdx` — the beta readiness review called this the one disclosure to fix before un-flagging, and until now the only disclaimer was "responses can be inaccurate".

Also reconciles the two feedback channels, which previously didn't know about each other: the Feedback button goes to the support form (quick, private, no account — the beta signal), while the forum and GitHub handle help and catalog gaps, which is already what the agent's own replies tell people. And retires stale copy — the FAQ still listed the assistant as a future feature "powered by open-source LLMs hosted at TACC" that searches NCBI, contradicting both the new disclosure and the documented inability to browse external resources; the roadmap still said "currently in testing".

## Verification

581 frontend tests pass, tsc and ESLint clean. Checked in a dev server that the nav item appears, `/assistant` renders instead of 404ing, the Learn card is back, the docs render, and the Feedback button points at the right form with `target="_blank"` and `rel="noopener noreferrer"`.

## Still open

- **#1290 also wants the Google Form renamed to "Help and Feedback."** Google-side, not code — and it matters more now that the docs describe that form as the feedback channel specifically.
- **Remaining Phase 1 issues:** #1293, #1299, #1318, #1320.
- **Worth knowing before this merges:** there's still no durable record of assistant conversations. #1294 is closed and the decision was durable-to-Postgres, but only the `saved_analyses` path landed, which needs an authenticated user who explicitly saves. Everything else lives in Redis for 2h and evaporates. So as it stands the beta can't answer "what are people asking it," which #1288 wanted in place first. Doesn't block this PR, but it does undercut the point of the beta.

### DIFF
--- a/app/views/AssistantView/assistantView.styles.ts
+++ b/app/views/AssistantView/assistantView.styles.ts
@@ -15,6 +15,13 @@ export const SectionContent = styled(Stack)`
   width: calc(100% - 32px);
 `;
 
+export const ActionsRow = styled(Box)({
+  alignItems: "center",
+  display: "flex",
+  justifyContent: "space-between",
+  paddingBottom: "8px",
+});
+
 export const TwoPanelLayout = styled(Box)({
   "@media (min-width: 960px)": {
     flexDirection: "row",
--- a/app/views/AssistantView/assistantView.tsx
+++ b/app/views/AssistantView/assistantView.tsx
@@ -1,13 +1,14 @@
 import { ChatPanel, SchemaPanel } from "@/components/Assistant";
+import { config } from "@/config/config";
 import { useAssistantChat } from "@/hooks/useAssistantChat";
 import { assistantAPIClient } from "@/services/assistant-api-client";
-import { useFeatureFlag } from "@databiosphere/findable-ui/lib/hooks/useFeatureFlag/useFeatureFlag";
+import FeedbackOutlinedIcon from "@mui/icons-material/FeedbackOutlined";
 import RestartAltIcon from "@mui/icons-material/RestartAlt";
 import { Button } from "@mui/material";
 import type { AssistantInfoResponse } from "@repo/shared/services/api-client/types";
-import Error from "next/error";
 import { type JSX, useEffect, useState } from "react";
 import {
+  ActionsRow,
   AssistantDisclaimer,
   ChatColumn,
   SchemaColumn,
@@ -22,7 +23,6 @@ interface Props {
 }
 
 export const AssistantView = ({ initialSessionId }: Props): JSX.Element => {
-  const isAssistantEnabled = useFeatureFlag("assistant");
   const {
     error,
     handoffUrl,
@@ -41,9 +41,11 @@ export const AssistantView = ({ initialSessionId }: Props): JSX.Element => {
     initialSessionId,
   });
   const [info, setInfo] = useState<AssistantInfoResponse | null>(null);
+  // Read per-site rather than importing one site's config, so the button can
+  // never point at another tenant's form.
+  const { supportUrl } = config();
 
   useEffect(() => {
-    if (!isAssistantEnabled) return;
     let cancelled = false;
     assistantAPIClient
       .assistantInfo()
@@ -56,9 +58,7 @@ export const AssistantView = ({ initialSessionId }: Props): JSX.Element => {
     return (): void => {
       cancelled = true;
     };
-  }, [isAssistantEnabled]);
-
-  if (!isAssistantEnabled) return <Error statusCode={404} />;
+  }, []);
 
   // On error there are no messages and no schema, which would hide Reset
   // exactly when it's the only way to clear a bad session id.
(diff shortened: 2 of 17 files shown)


NoopDog (approved):

frano-m:
Rebased onto current `main` (single commit, original authorship preserved). Both site builds green; tsc/eslint/prettier clean. All the assistant/Feedback/flag-removal changes are re-applied at their post-split locations (`sites/brc-analytics/…`, `@repo/shared/…`).

**Two deliberate changes to the learn gating** — flagging since they diverge from the original diff:

1. **`pages/learn/[...slug].tsx` — kept `main`'s version; dropped the `SITE_SPECIFIC_DOCS`/`isDocAllowed` mechanism.** That was written when `/learn` was a shared tree. On `main` the whole route is now gated on `ROUTES.LEARN` and the docs live under `sites/brc-analytics/docs/`, so GA2 already builds **zero** learn pages — the per-doc assistant filter is redundant. The end goal (no assistant doc reachable on GA2) is still met, just more broadly.

2. **`LearnView` — removed the `assistant` flag gate (the PR's intent) but didn't carry over the `allowedPaths.includes(ROUTES.ASSISTANT)` card check.** LearnView is BRC-only post-split and BRC sets no `allowedPaths`, so that check is always true; the assistant card simply always shows on BRC.

**Kept as-is:** the `pages/assistant/{index,saved}.tsx` `notFound` guards on `ROUTES.ASSISTANT` — still needed since both sites build from the shared `pages/` tree, and it's the same route-gate pattern as our learn gate.

PR #1567, opened 2026-07-30
refactor: move sectionhero cluster into @repo/shared (#1561)

frano-m:
## Summary

Moves the shared `SectionHero` into `@repo/shared` as a self-contained cluster, so any site resolves it from a single site-neutral home.

- `git mv` the generic `SectionHero` (`sectionHero.tsx` + styles + its `Hero`) → `packages/shared/components/layout/SectionHero/`; repointed all consumers.
- Folded the previously-standalone `Hero` primitives (`common/{constants,utils}`, `components/Defs/{SmokeCircle,SmokeRect}`) into the cluster under `components/Hero/`, and the `useResizeObserver` render-prop `Section` under `components/Section/` — the cluster is now self-contained.
- Renamed `Hero/common/entities.ts` → `Hero/common/types.ts`; extracted the inline component prop interfaces into co-located `types.ts` files.
- Repointed the Home section hero (stays app-side) to import the primitives from the shared cluster.

Pure reorg — no behaviour change.

Closes #1561

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1569, opened 2026-07-30
refactor: move docs sectioncontent and content model into @repo/shared (#1568)

### LINKED ISSUE #1568: refactor: move docs SectionContent and content model into @repo/shared
Move the Docs `SectionContent` component and its supporting content model into `packages/shared/views/docs` so the shared content views can consume them without importing from `@/`.

Closure moved (all site-neutral):
- `SectionContent` (component + styles)
- `content.styles` (+ `headline.styles`) — the shared content-page styling
- `docs/common/frontmatter/types` and `docs/common/staticGeneration/types` — the content frontmatter / static-generation model
- `SectionContentCards/types` and `SectionOverview/types` — type-only, referenced by the frontmatter model (their components stay app-side for now)

Prerequisite for the shared content index view. Part of the site-separation shared-extraction work (#1246 / #1417).

frano-m:
## Summary

Moves the Docs `SectionContent` component and its supporting content model into `packages/shared/views/docs`, so shared content views can consume them without importing from `@/`.

Moved (all site-neutral, closure of `SectionContent`):
- `SectionContent` (component + styles) → `views/docs/components/SectionContent/`
- `content.styles` (+ `headline.styles`) → `views/docs/` — the shared content-page styling
- `docs/common/{frontmatter,staticGeneration}/types` → `views/docs/common/` — the content frontmatter / static-generation model
- `SectionContentCards/types` + `SectionOverview/types` → `views/docs/components/…` — type-only; referenced by the frontmatter model (their components stay app-side for now)

Consumers repointed; `packages/shared/views/docs` verified free of `@/` and `@site-config` imports.

Pure reorg — no behaviour change.

Closes #1568


NoopDog:
**Nit — normalize intra-package self-imports to relative (same as #1567)**

The new `views/docs` cluster uses absolute `@repo/shared/views/docs/...` paths for several of its own internal imports (in `sectionContent.tsx`, `sectionContent.styles.ts`, and `common/frontmatter/types.ts`). Each of these is exactly `../../`-depth relative, i.e. within the cap we settled on in #1567, so by that same convention they should be relative:

```diff
# components/SectionContent/sectionContent.tsx
-import type { StaticProps } from "@repo/shared/views/docs/common/staticGeneration/types";
-import { Section } from "@repo/shared/views/docs/content.styles";
+import type { StaticProps } from "../../common/staticGeneration/types";
+import { Section } from "../../content.styles";

# components/SectionContent/sectionContent.styles.ts
-import type { FrontmatterProps } from "@repo/shared/views/docs/common/frontmatter/types";
-import { CONTENT_TYPE } from "@repo/shared/views/docs/common/frontmatter/types";
+import type { FrontmatterProps } from "../../common/frontmatter/types";
+import { CONTENT_TYPE } from "../../common/frontmatter/types";

# common/frontmatter/types.ts
-import type { CardItem } from "@repo/shared/views/docs/components/SectionContentCards/types";
-import type { Overview } from "@repo/shared/views/docs/components/SectionOverview/types";
+import type { CardItem } from "../../components/SectionContentCards/types";
+import type { Overview } from "../../components/SectionOverview/types";
```

Non-blocking (tsc is green) — just for consistency with #1567 and the rest of `@repo/shared`, which uses relative self-imports for anything ≤ `../../`.

frano-m:
Thanks — leaving these as the `@repo/shared` alias for now. These are all exactly `../../`-depth (2 parents), and CODING_STANDARDS says relative only for same-dir (`./x`) and single-parent (`../x`), alias for anything 2+ `../` — so the alias is the standards-correct form here. #1567 is consistent with that too (its only relatives are `./`-descendant or single `../`; the 3-level Defs→common refs are already aliased).

That said, `@repo/shared` is genuinely mixed today (some absolute self-imports, some relative up to `../../../`), so rather than flip these piecemeal we'll do a dedicated sweep later to land one unified self-import convention across the package. Tracking it as non-blocking for this PR.

dannon:
+1 to doing this as one sweep rather than piecemeal. Let me throw in a vote on what the rule should be, though, since I think the depth cap is what's causing the friction here.

The problem with capping at `../../` is that the rule's answer depends on how deep a file happens to sit rather than on anything about the import itself.

`sectionContent.tsx` lives at `views/docs/components/SectionContent/` and imports `views/docs/common/staticGeneration/types`. Up to `components/`, up to `docs/`, back down -- `../../common/staticGeneration/types`, exactly 2 parents, inside the cap. So relative is correct today.

Now nest `SectionContent` under a parent component, which is our default move -- the `X/components/Y/` pattern shows up four levels deep in `Table/components/TableCell/components/SpeciesCell/components/TagList/`. Say it becomes `views/docs/components/ContentLayout/components/SectionContent/`. Recount to the same target: up to ContentLayout's `components/`, up to `ContentLayout/`, up to docs' `components/`, up to `docs/`, back down. That's `../../../../` -- over the cap, so the line has to be rewritten as `@repo/shared/views/docs/common/staticGeneration/types`.

The file didn't change and the target didn't move. The relationship is identical -- reach the cluster's shared types. But the rule gives a different answer, purely because something upstream got one level deeper.

Two problems with that. A PR whose actual change is "nest SectionContent under ContentLayout" now also rewrites imports in files whose dependencies didn't change, in a diff that's already mostly import churn. And you can't verify `../../common/staticGeneration/types` by reading it -- you need the importing file's own depth and have to do the arithmetic. Which is what happened across #1567 and #1569: two people counted, got different answers about what the rule allowed, and neither could point at anything authoritative.

What I'd like us to standardize on is what we do in Galaxy: `./` for same-dir and descendants, `@repo/shared/...` for everything else, no `../` ever. One rule, no arithmetic, and it doesn't move when files do.

The one real argument for relative self-imports is that cluster moves get cheaper -- `git mv` a directory and nothing inside it changes. I don't think that survives the numbers, though. This PR is 39 added lines and 36 of them are imports, because repointing consumers dominates. Making the 5 internal ones relative saves 5 lines out of 36 and costs us a second rule.

What absolute buys us that I actually care about right now: internal and external references to a shared module are the same string. `grep '@repo/shared/views/docs/content.styles'` finds every consumer in one shot. In the middle of an extraction where we're constantly asking "who still depends on this," that's worth more than the churn.

Downside, to be fair -- co-located types get long. `TagList/tagList.tsx` importing `../../types` becomes `@repo/shared/components/Table/components/TableCell/components/SpeciesCell/types`, and there are about 5 of those today. I'd eat it, since prettier and organize-imports mean nobody's hand-maintaining those lines. If the length ever becomes a real problem the fix is a barrel at the cluster root for consumers, not a relative exception.

Assuming no objections, let's make it a lint rule so it stops being a review conversation. `eslint-plugin-import` isn't installed, but `no-restricted-imports` covers it:

```js
{
  files: ["packages/shared/**"],
  rules: {
    "no-restricted-imports": ["error", {
      patterns: [{
        group: ["../**"],
        message: "Use @repo/shared/... -- relative imports only for ./ descendants.",
      }],
    }],
  },
}
```

@frano-m side note -- where does `CODING_STANDARDS` live? I couldn't find it in the repo. Might be worth checking a short version in alongside the lint rule so the convention and the enforcement are in the same place.

NoopDog (approved):

frano-m:
@dannon Totally agree!

And yes to the lint rule so we stop relitigating it in review — I'll spin up a follow-up ticket to set it up (prettier + eslint), covering `@repo/shared` and both site packages.

Sorry about the `CODING_STANDARDS` reference — it's a personal rules file of mine, and I didn't catch that Claude had cited it in the PR.

Thanks for thinking this through — much better place to land. 🙏

Follow-up ticket: #1573

PR #1631, opened 2026-08-05
refactor: relocate assistant, stepper, viewmodelbuilders — batch C (#1616, #1617, #1618, #1619)

frano-m:
## Summary

Third site-separation relocation batch (mechanical `git mv` + import repoints, no behaviour change). Batched per NoopDog's request; each ticket preserved as its own commit.

- **#1616** — relocate the assistant into the shared `AssistantView`
- **#1617** — decouple the workflow stepper cluster from app viewModelBuilders + site unions
- **#1618** — relocate the workflow stepper cluster into the shared package
- **#1619** — relocate the per-site viewModelBuilders into the sites, and promote the shared assembly builders into `@repo/shared`

Closes #1616
Closes #1617
Closes #1618
Closes #1619

Part of the site-separation epic (#1246).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #1629, opened 2026-08-05
refactor: resolve the assembly side column from entity config instead of appKey dispatch


NoopDog:
## Background

PR #1628 (part of the site-separation epic #1246) relocated `AssemblyView` and `AnalyzeWorkflowsView` into `packages/shared` and dropped the runtime `Side` dispatcher. Because the shared package must not import site-specific components, each call site now selects the side column imperatively:

```tsx
config().appKey === APP_KEYS.GA2
  ? <AssemblyView SideComponent={GA2Side} entityId={...} />
  : <AssemblyView SideComponent={BRCSide} entityId={...} />
```

This same `appKey`-ternary + dual (`GA2Side`, `BRCSide`) imports are duplicated in two files:
- `app/views/AnalyzeWorkflowsRouteView/analyzeWorkflowsRouteView.tsx`
- `pages/data/[entityListType]/[entityId]/index.tsx`

It works and is type-safe as written, but every new shared assembly view repeats the ternary, and every new site adds a branch to each existing ternary — the two copies will drift.

## Proposal

Stop dispatching on `appKey` at the call site and resolve the assembly side column from the entity config, the way `EntityView` already renders `sideColumn` / `top` via `ComponentCreator` with **zero** `appKey` branching (`packages/shared/views/EntityView/entityView.tsx`). Each site already supplies its own entity config (selected per-site at runtime), so:

- The BRC and GA2 assembly entity configs each carry their own assembly side component (declaratively, like `priorityPathogenSideColumn`).
- The shared `AssemblyView` / `AnalyzeWorkflowsView` resolve it from config once.
- A third site becomes a config-only change with no call-site edits, and the duplicated imports/ternaries go away.

## Acceptance criteria

- [ ] `app/views/AnalyzeWorkflowsRouteView/analyzeWorkflowsRouteView.tsx` and `pages/data/[entityListType]/[entityId]/index.tsx` no longer branch on `appKey` to choose the side component.
- [ ] The assembly side column is supplied via entity config and resolved in the shared view.
- [ ] Typecheck and existing tests pass; no visual/behavioral change for BRC or GA2.

## Notes

- Keep the `<T extends AssemblyContract>` generic on the shared views — it is load-bearing (the site Side components take narrower assembly types than `AssemblyContract`).
- Surfaced during review of #1628. Non-blocking; that PR landed with the inline dispatch intentionally.

Part of #1246.

frano-m:
[...]

Issue #1417, opened 2026-07-08
[GA2/BRC split] 5: Initialise workspace + per-site Next.js apps (absorbs Issue 3)


frano-m:
Part of #1246 · **Phase 2 (COMMITTED — gate resolved GO 2026-07-16)** · Bucket: rare-window · Size: L

> **Go decision (2026-07-16 team sync, NoopDog + fran):** proceed with the full workspace split. Confirmed requirements: full clean folder/`pages/` separation (fran) and **independent deploy — separate apps** (NoopDog). See `.claude/rules/MONOREPO_PLAN.md` §0.

## Absorbs Issue 3 (#1415, closed NOT_PLANNED)
Removing the `appKey`/`allowedPaths` branching was deferred into this ticket, because splitting `pages/` per-site is what actually eliminates the branches — doing it standalone would only add throwaway config plumbing. As of 2026-07-16 that branching is still fully live and must be removed here:
- **45 `appKey`/`APP_KEYS` usages across 13 files** — `app/common/meta/utils.ts` + `pages/index.tsx`, `pages/_app.tsx`, `pages/about/*`, and all `pages/data/*`.
- **`allowedPaths` in 6 pages** (`calendar`, `roadmap`, `about`, `about/roadmap`, `about/vision`, `about/partner-resources`) + `site-config/common/entities.ts` + `site-config/ga2/local/config.ts`.

## Do this first (not part of this ticket, but a prerequisite)
- Upgrade `@databiosphere/findable-ui` 54 → 55 on a clean base (own PR), land and verify green before starting the split.

## Scope
Set up npm/pnpm workspaces. Create `packages/core/`. Create `sites/brc-analytics/` and `sites/ga2/` as independent Next.js apps, each with its own `pages/`, `next.config.mjs`, `public/`, `.env`. Site-specific components/providers/utils/types live under `sites/<site>/`. Remove the `appKey`/`allowedPaths` branching (per "Absorbs Issue 3" above), the `NEXT_PUBLIC_SITE_CONFIG` config map, and the `appKey` / `allowedPaths` types. Pick a low-activity window; pre-flight comms; expect concentrated work (the 1–2 day estimate predates the Issue 3 fold-in — budget more).

## Acceptance criteria
- [ ] Both sites build and pass tests as independent workspace apps.
- [ ] Shared code in `packages/core` has no site-specific imports.
- [ ] `app/config/config.ts` config map removed; each site imports its own config directly.
[...]

PR #1633, opened 2026-08-05
refactor: explicit assembly and organism analyze-workflows pages (#1621)

### LINKED ISSUE #1621: refactor: explicit assembly and organism analyze-workflows pages
Split the single dynamic `data/[entityListType]/[entityId]/analyze/workflows` route into explicit `assemblies` and `organisms` pages, dissolving the runtime `entityListType` / redirect / list-vs-configure branching in `AnalyzeWorkflowsRouteView` (which is removed).

Each explicit page owns its own flow:
- **assembly**: `trsId` → configure inputs; no `trsId` → compatible-workflows list.
- **organism**: `trsId` → configure inputs; no `trsId` → redirect to the organism detail page.

This replaces the catch-all dynamic segment with static per-type routes — groundwork for the per-site pages split. A small entity static-paths helper is kept app-level for now; the shared `staticGeneration` extraction is deferred to the pages split.

Part of the site-separation epic #1246.

frano-m:
## Summary

Replaces the dynamic `data/[entityListType]/[entityId]/analyze/workflows` route (and the `AnalyzeWorkflowsRouteView` dispatcher) with **explicit per-type pages**, each rendering its concrete view directly:

- **`data/assemblies/[entityId]/analyze/workflows`** — compatible-workflows list, or `WorkflowInputsView` when a `trsId` is selected.
- **`data/organisms/[entityId]/analyze/workflows`** — `OrganismWorkflowInputsView`; a bare URL (no `trsId`) redirects to the organism detail page.

Also extracts the shared SSG helpers into `@repo/shared/services/staticGeneration` (`makeEntityStaticPaths`, `EntityPageParams` / `EntityPageProps`) and relocates `entityUtils` / `seedDatabase` into the shared package.

**Priority-pathogen chip preserved:** each explicit page injects the per-site `organismBuilder` by appKey (BRC → priority-pathogen chip, GA2 → shared default), matching the wiring the dissolved `AnalyzeWorkflowsRouteView` carried.

Closes #1621


NoopDog (approved):
Approving. ✅

Reviewed for correctness (routing being the main risk, since typecheck/Jest don't cover it). Verified green: typecheck clean, **599/599 tests pass**, no dangling references to the deleted dispatcher or old util paths.

**What I checked:**
- **Behavior parity is exact.** The assembly page (list when no `trsId`, `WorkflowInputsView` when selected) and organism page (`OrganismWorkflowInputsView` with `trsId`, redirect to the organism detail on a bare URL) reproduce the dissolved `AnalyzeWorkflowsRouteView` branch-for-branch, including the empty-`?trsId=` handling and `isReady` guards.
- **No Next.js route conflict.** Enumerated the `pages/data` tree: at depth 2 the single dynamic slug `[entityListType]` coexists with static siblings (`assemblies`, `organisms`, …) — the supported "static wins" pattern — and `[entityId]`/`[trsId]` never collide as differing slug names at the same depth. The static `assemblies`/`organisms` subtrees only define `…/analyze/workflows`, so detail (`/data/assemblies/<id>`) and `…/analyze/custom` still fall through to the dynamic routes; the old dynamic `analyze/workflows` file is deleted, so no two pages emit the same static path. URLs unchanged. (Definitive route-conflict validation is `next build` in CI — my analysis is static but found none.)
- **Priority-pathogen chip genuinely preserved — this actually resolves the #1631 finding.** `SideColumn` now takes `organismBuilder = defaultOrganismBuilder` (shared, no-chip) and renders `{...organismBuilder(organism)}`; both views thread the prop through. Each page injects `buildBRCOrganismDetails` for BRC (chip restored) or `undefined` for GA2 (falls back to the shared default — passing `undefined` correctly triggers the default param). Nice.
- **SSG extraction is faithful** — `makeEntityStaticPaths` emits one path per entity, equivalent to the old filtered loop, with config injected so shared stays decoupled.

**Recommended follow-up (non-blocking):**
The dissolved dispatcher's logic is now duplicated across the two new pages — `trsId` parsing, the `isReady` guard, and especially `organismBuilder = isGA2 ? undefined : buildBRCOrganismDetails` (the appKey→builder decision went from one site to two). This is the same call-site component/builder injection pattern already tracked in #1629. Suggest a small shared helper (e.g. `getOrganismBuilder(appKey)` and/or a `useTrsIdFromQuery()` hook) to DRY the shared bits while keeping the per-page `getStaticPaths`/`getStaticProps` that Next requires — or fold it into whatever comes out of #1629. Happy to file it if you'd like.

Minor: the assembly page computes both `workflowsList` and `organismBuilder` each render though only one path is used (negligible); `EntityPageProps<never>` to signal "no build-time data" is slightly unusual but the doc comment makes it clear.

Great cleanup overall — swaps the runtime dispatcher for explicit pages with identical behaviour and URLs, and fixes the #1631 chip regression as a bonus.

frano-m:
Thanks @NoopDog — thorough routing/route-conflict check, that was the real risk here. Also rebased onto `main` now that #1632 (batch D) landed — clean (D is theme-only, E is pages/services, disjoint), re-verified 599/599 + both builds green.

**Follow-up (duplication) — leaving, points at #1417.** The `organismBuilder = isGA2 ? undefined : buildBRCOrganismDetails` branch is exactly the call-site injection pattern from #1629, which we closed as covered by #1417. At the per-site `pages/` split each site's own pages resolve their builder directly (BRC hardcodes `buildBRCOrganismDetails`, GA2 omits it → shared default), so the appKey→builder decision *dissolves* there. Extracting `getOrganismBuilder(appKey)` now would be throwaway plumbing #1417 deletes, so I'll leave it.

The `trsId`-parse + `isReady` guard is the part that *survives* the split (both sites keep it), so `useTrsIdFromQuery()` is a fair standalone DRY — but it's ~3 lines with the analyze pages as its only consumers today, so I'd rather fold it into #1417's per-site page work than add a hook now. No new ticket unless you'd prefer one.

**Minors — leaving:** the assembly page computing both `workflowsList` and `organismBuilder` each render is negligible (both cheap, and it drops out per-site at #1417); `EntityPageProps<never>` is intentional ("no build-time props beyond meta") and the doc comment calls it out.

PR #1634, opened 2026-08-05
feat: log assistant turns durably for beta review

### LINKED ISSUE #1294: Assistant: verify and enable logging of user queries and assistant responses

dannon:
Closes #1294.

Today's coverage is partial: #1301 persists the conversation to `saved_analyses` when someone explicitly saves one, and token counts and run latency go to the app logs. The gap is the sessions nobody saves, which is most of them -- those live only in Redis on a 2h TTL, and Redis runs `allkeys-lru` at 256MB shared with caches, so they can evict earlier than that.

## Summary

- New `assistant_turn_log` table, one row per turn: the message, the reply, that turn's tool calls and returns, the tracker snapshot, token counts, latency, and the model that produced it.
- A sibling table rather than reusing `saved_analyses` as originally discussed. `saved_analyses.user_id` is `nullable=False, ON DELETE CASCADE`, so reusing it means loosening both on a live user-facing table, after which `list_saved_analyses_for_user` / `get_saved_analysis` / `delete_saved_analysis` all have to keep their filters right forever. One slip puts a stranger's conversation in somebody's saved list. Modeled on `workflow_runs` instead (nullable `user_id`, `ON DELETE SET NULL`).
- Failed turns get a row too (`outcome='error'`, exception class in `error_kind`, prompt kept). Otherwise the corpus is only the requests that happened to work, which is backwards for a beta.
- Every turn carries a `turn_id` that's also set as a Sentry tag (`assistant.turn_id`), so an exception there joins back to the prompt that caused it.
- Retention is 90 days, swept by the app on a timer (`app/services/turn_log_retention.py`) rather than a cron job someone has to remember to install. The purge is an idempotent `DELETE` by age, so running it in several workers is harmless; `scripts/purge_assistant_turn_logs.py` stays for manual use.
- User notice on the assistant page and in `/learn/assistant`.

## Notes for review

**The write can't cost a user their reply.** It runs off the response path as a detached task, fail-open, with failures reported to Sentry so a quietly incomplete corpus doesn't look like a quiet beta.

It's deliberately **not** a `Depends(get_db_session)`: that raises when `DATABASE_URL` is unset and dependencies resolve before the handler body, so it would 500 every chat in the default `backend/docker-compose.yml`, which sets no `DATABASE_URL`. It's also not a FastAPI `BackgroundTask` -- those attach to a returned response and would skip exactly the error paths we now want recorded.

**Transcripts are capped** at 64KB (`ASSISTANT_TURN_LOG_MAX_TRANSCRIPT_BYTES`). Tool returns are unbounded and one broad `query_catalog` call would otherwise land in the row, the WAL, and every backup. Messages are kept from the front so the prompt and first tool calls always survive; `transcript_truncated` marks rows where it bit.

**`turn_index` is an ordering hint, not a sequence.** It's read from Redis session metadata, so two concurrent requests on one session can share a value. `created_at` orders rows and `turn_id` is the unique handle -- there's a test asserting a collision doesn't cost us a row, since a unique constraint would silently drop one under the fail-open write.

Migration `0003` is amended in place rather than stacked, since it hasn't been deployed anywhere.

## Test plan

- [x] `pytest` -- 494 pass (`test_links.py` needs a live Docker backend on :8080 and fails on main too)
- [x] Migration upgrade and downgrade verified on sqlite; no autogenerate drift for the new table
- [x] App lifespan smoke-tested: purge task starts, runs, and cancels cleanly on shutdown
- [x] `ruff format --check`, prettier on the touched frontend files
- [ ] Confirm `DATABASE_URL` is set wherever the assistant is deployed -- with no DB nothing is recorded and the app only warns at startup
- [ ] Align DB backup retention with the 90-day disclosure in `/learn/assistant`

## Follow-ups

- `.env.example` needs the new settings added by hand; they're documented in `backend/api/README.md` for now.
- `instrument=True` on pydantic-ai would give per-tool span waterfalls in Sentry, which a single `latency_ms` column can't. Complementary to this, but wants a check that prompts don't leak into span content first.
- #1293 can read funnel counts off this table now. Example review queries are in the backend README.


NoopDog:
Ran a deep review pass over this PR (code review + security review + quality sweep, plus adjudicating Copilot's suppressed comments against HEAD). This is a carefully built PR — the fail-open write path, drain-before-shutdown ordering, tz-aware cutoff math, indexes, and migration all check out, and all 36 new tests pass locally. A handful of things worth a look before merge, one of them the clear headline.

## The one to fix before merge

**Logging and disclosure are gated on different predicates.** `turn_log.schedule()` writes whenever `ASSISTANT_TURN_LOGGING_ENABLED && DATABASE_URL`, but `/assistant/info` only advertises retention when purge is *also* enabled and retention ≥ 1. So with `ASSISTANT_TURN_LOG_PURGE_ENABLED=false` (or retention ≤ 0), every turn is logged **with no deletion**, the UI notice disappears — and `assistant.mdx` promises *"If the notice isn't there, nothing is being kept."* Defaults are safe, so this only bites on misconfiguration, but the failure mode is inverted from safe: log more, disclose less. Suggest one predicate owned by `turn_log` (e.g. `advertised_retention_days()`) shared by `schedule()`, the sweep, `/info`, and the boot warning — which currently says "despite the notice shown to users" in exactly the config where no notice is shown.

## Bugs

1. **A failing `on_turn` sink costs the user their reply** — `assistant_agent.py:~1215`: the success-path `on_turn(telemetry)` runs inside the `try`, so a raising sink (e.g. `create_task` on a shutting-down loop) discards the computed reply, records a bogus `outcome='error'` row, and returns a 500 — the fail-open contract unenforced at its own boundary. Wrap both `on_turn` calls in `try/except`.
2. **Error telemetry can attach to an unvalidated session** — `assistant_agent.py:~1037`: `progress.get("session_id", session_id)` falls back to the raw request-supplied id when `require_session` fails with something besides `KeyError`/`PermissionError` (e.g. Redis `ConnectionError`) — bypassing the PermissionError carve-out. `progress.get("session_id")` alone is honest.
3. **Endpoint tests aren't hermetic** — `tests/conftest.py`: `app_with_stubbed_agent` doesn't pin `DATABASE_URL`, `config.py` runs `load_dotenv()`, and CI copies `.env.example` (uncommented `DATABASE_URL`) to `.env` — so the chat tests schedule real detached writes in CI, and on a dev box with a live DB they insert junk rows. One test defensively delenvs; the rest don't. Fix: `monkeypatch.delenv("DATABASE_URL", raising=False)` in the fixture.
4. **Failed writes leak conversation text to Sentry/logs** — `turn_log.py:~120`: SQLAlchemy exception strings include bound parameters (message/reply/transcript fragments), and sentry-sdk defaults `include_local_variables=True` with no scrubbing configured — so a transient DB failure copies conversation content into Sentry, outside the 90-day promise. The timeout branch already does this right (class name + turn_id only); suggest matching it, plus `hide_parameters=True` on the engine.

Smaller: `_strip_nuls` skips dict *keys* (a NUL key still kills the JSONB insert — one-token fix); `turn_log_retention_days` is non-optional in the TS type, so a frontend deployed ahead of the backend violates it (make it `?:`); the ORM declares `JSON` where the migration creates `JSONB`; `assistant.mdx` hardcodes "90 days" in prose for a window that's env-configurable and already served dynamically via `/info`.

## Copilot's suppressed comments, adjudicated

Copilot ran 7 reviews and GitHub auto-suppressed most of the feedback — for the record: **12 of 15 distinct findings are already fixed at HEAD** (bytes-vs-chars cap, shutdown drain, monotonic clocks, config duplicates, error-path session attribution, `/info` gating, and more). The 3 still open are all in `turn_log.py`: the predicate mismatch above, the NUL-key nit, and the stale warning wording.

## Quality suggestions (take or leave)

- `chat_with_telemetry` publishes telemetry twice — via `on_turn` *and* a returned tuple every production caller discards. Keeping only the callback lets `chat()`/`chat_with_telemetry()` collapse into one method, and the `progress` out-param dict could become a single mutable draft-telemetry object.
- The transcript is JSON-serialized on the response path even when logging is disabled, and **twice** when over budget — the exact multi-MB `query_catalog` case the cap exists for stalls the event loop. A single incremental pass with early exit fixes both; capping could also move into the detached writer.
- The retention sweep purges immediately at boot in every worker with one unbounded `DELETE` — sleep-first (or jitter) plus batching would be kinder to the DB.
- Test dedup: the sqlite factory, `_stub_agent`, and the `_fake_session` monkeypatch (×4) duplicate fixtures this PR's own `conftest.py` was created to share.

Verified clean: no endpoint exposes turn-log rows, all SQL parameterized, no cross-user attribution under concurrency, Sentry tag scoping per-request, migration/downgrade correct, README queries match the schema and indexes.

---

NoopDog:
Re-reviewed after `c3b761c4` — the predicate unification is exactly right: `active_retention_days()` as the single source of truth, fail-safe coupling (sweep off → logging off), and the parametrized tests pin both halves across all five off-configurations. That also makes the mdx "if the notice isn't there, nothing is being kept" sentence actually true. Nice find on the `_pending_writes` test leak too.

Four items from the earlier review are still open, plus one suppressed Copilot comment worth taking. Ranked:

## Still open

**1. Failed turn-log writes leak conversation text to Sentry/logs** — `turn_log.py` `except Exception` branch (~line 138). `logger.exception` appends the traceback, and SQLAlchemy exception strings embed the bound parameters (`user_message`, `assistant_reply`, transcript fragments); bare `sentry_sdk.capture_exception()` additionally ships frame locals (the `telemetry` object) since sentry-sdk defaults `include_local_variables=True` and the init sets no scrubbing. Any transient DB failure during the beta copies conversation content into Sentry — outside the 90-day promise, in a third-party store. The adjacent timeout branch already does this right (ids only + `capture_message`); suggest matching it, plus `hide_parameters=True` on the engine and/or `include_local_variables=False` in the Sentry init. Given the PR's privacy framing, this is the one I'd fix before merge.

**2. A raising `on_turn` sink still costs the user their reply** — `assistant_agent.py:1216`. The success-path `on_turn(telemetry)` runs inside the `try`, so a sink failure (e.g. `create_task` on a shutting-down loop) discards the computed reply, records a bogus `outcome='error'` row, and returns a 500; on the error path a sink failure replaces the original exception. One `try/except Exception: logger.exception(...)` around each of the two invocations enforces the fail-open contract at its own boundary.

**3. Error telemetry can attach to an unvalidated session** — `assistant_agent.py:1037`. `progress.get("session_id", session_id)` falls back to the raw request-supplied id when `require_session` fails with something other than `KeyError`/`PermissionError` (e.g. Redis `ConnectionError`) — bypassing the PermissionError carve-out that exists to keep a stranger's message out of someone else's session. `progress.get("session_id")` alone is honest: null means no validated session.

**4. Endpoint tests still schedule real detached DB writes** — `tests/conftest.py` `app_with_stubbed_agent` doesn't clear `DATABASE_URL`; `config.py` runs `load_dotenv()` and CI copies `.env.example` (which has an uncommented `DATABASE_URL`) to `.env`, so under default all-on settings every successful `POST /chat` in the endpoint tests calls the real `schedule()`. The new `_no_leaked_turn_log_writes` fixture fixes the cross-test blocking symptom but not the writes. `monkeypatch.delenv("DATABASE_URL", raising=False)` in the fixture makes the default hermetic; tests that want scheduling opt in.

## The suppressed Copilot comment (valid, minor)

`start_purge_task()` (~line 232) still starts the sweep loop when `RETENTION_DAYS < 1` — it then spins forever with `purge_expired()` refusing and logging an error every interval. Now that `active_retention_days()` exists, the fix is one more caller: return early (with the warning) when it's `None`. That also collapses the remaining hand-rolled conditions in that function onto the single predicate.

## Smaller / take-or-leave

- `packages/shared/services/api-client/types.ts:62` — `turn_log_retention_days` is non-optional, so a frontend deployed ahead of the backend violates the type (field absent → `undefined`). `?: number | null` covers rollout skew.
- ORM declares `JSON` where the migration creates `JSONB` — latent `create_all`/autogenerate drift; `JSON().with_variant(JSONB, "postgresql")` aligns them.
- assistant.mdx still hardcodes "on this site it's 90 days" in prose for an env-configurable window the page now reports dynamically — dropping the number ends the last doc-drift risk.

NoopDog:
One design thought (fine as a post-merge follow-up too): I think we may be able to simplify the write path in a way that trims a good chunk of code — and makes most of the open review items disappear as a side effect.

"A log write must never cost a user their reply" is the right contract — but it's the `try/except` that protects the reply, not the backgrounding, and a turn takes seconds of inference while the insert takes milliseconds. Three simpler shapes, cheapest first:

- **Inline fail-open write** — `try: await insert / except: log + Sentry` just before returning; error turns write in the `except` handler where the exception is already in hand. Same guarantee, ~0.1% latency cost, zero task machinery. *Idiomatic when the write is fast and small relative to the request — which is this case; my recommendation.*
- **`BackgroundTasks` for success + inline write for errors** — success rows go through FastAPI's built-in post-response hook (framework-managed lifecycle, truly zero added latency); the error path writes inline, which answers your notes' objection that error responses skip background tasks. *Idiomatic when you want post-response execution without owning any task lifecycle — the FastAPI-native middle ground.*
- **One writer task + bounded `asyncio.Queue`** — `schedule()` becomes `put_nowait`; a single long-lived consumer owns all writes. One task to supervise and drain instead of N, backpressure for free. *Idiomatic when detachment is genuinely required (bursty writes, slow sink) — the standard asyncio shape for fire-and-forget.*

Any of the three dissolves most of the current surface — the `_pending_writes` set, semaphore, per-write timeouts, drain-before-`close_db()`, the test-leak fixture — and with it three of the four open review items. `db_session()` usage stays as-is, so the no-`DATABASE_URL` deployment still works in all of them.

The current design works and is well-tested — this is purely about carrying less machinery going forward. Happy to sketch the diff. Great feature for the beta either way. 🙌

dannon:
Backed the write-task machinery out of this and went back to an inline write, which is what it was originally. Worth writing down why, because the process went wrong in an instructive way.

The first version awaited the insert at the end of the request -- `try` / `wait_for` / `except` -> log, fail-open. Simple, and correct.

Review pointed out that a slow database would then add its timeout to every reply. True. So I moved the write to a detached task. That needed a set to hold task references; a drain before `close_db()` so shutdown didn't dispose the engine underneath in-flight writes; a semaphore so analytics writes couldn't starve the request pool; a `lifecycle()` context manager to order all of it; and an autouse test fixture once the module-global set started leaking between tests. Each one exists to prop up the one before it.

They also produced four defects that later passes then found: the semaphore sat outside the timeout, so it stopped bounding anything; the drain abandoned stragglers that rebuilt the engine after shutdown; the test leak took the suite from 6s to 101s; and a sink that raised could destroy a reply we'd already computed -- possible only because `create_task` can fail on a closing loop.

The problem all that bought us was worth about two seconds, in a scenario (Postgres unresponsive on the same host) about as likely as Redis failing, which we don't defend against anywhere else in this app. The insert is single-digit milliseconds against a turn that spends seconds in model inference. Inline also removes shutdown loss outright -- nothing is ever in flight, so there's nothing to drain.

The reviews weren't wrong on the facts; the latency observation was correct. What they can't do is weigh it. An automated reviewer reports a 2-second worst case on a rate-limited beta endpoint in the same tone and the same list position as a real privacy bug, and if you act on all of it you accumulate machinery per finding. Across Copilot's nine passes, a 54-agent review I ran myself, and the generated reviews on this PR, most of what the last three rounds found was created by fixes from the round before.

In fairness they did catch things I'd have shipped -- a 403 path that wrote the caller's message under the victim's `session_id`, a NUL byte silently dropping rows, a 90-day retention promise with nothing enforcing it, duplicate tests that never ran. That's a good hit rate on local defects. The failure was letting the same tooling drive an architecture decision, where it has no sense of proportion and I didn't supply one.

So: inline write, and `_pending_writes`, `drain()`, the semaphore and the lifecycle drain ordering all come out. 60 lines lighter, CI green.

This records what people type into a chat box during a beta. It should be about forty lines.

---

🤖 Summary from Claude Code -- which also wrote the machinery I just deleted, and the reviews that argued for it. The tool isn't the problem; not weighing what it says is.

NoopDog:
Re-reviewed at HEAD — the inline rewrite reads great, and everything from the earlier rounds checks out as fixed (predicate gates verified at all three consumers, session attribution, the sink guards, purge gating). Two small items and then this looks ready to merge from my side:

**1. The `except Exception` branch in `record()` can leak conversation text to logs/Sentry** (`turn_log.py:~100`). `logger.exception` includes the exception string, and SQLAlchemy embeds the bound parameters in it (`user_message`, `assistant_reply`, transcript fragments); bare `sentry_sdk.capture_exception()` additionally ships frame locals (the `telemetry` object) under sentry-sdk's default `include_local_variables=True`. So a routine DB error copies conversation content into app logs and Sentry — outside the retention window the PR enforces. The timeout branch directly above already has the right shape; suggest matching it:

```python
    except Exception:
        logger.warning(
            "Turn log write failed (turn_id=%s session=%s)",
            telemetry.turn_id,
            telemetry.session_id,
        )
        sentry_sdk.capture_message("Assistant turn log write failed", level="error")
```

(`hide_parameters=True` on the engine is a good belt-and-braces if you want the exception class preserved in Sentry instead.)

**2. Stale comment on the timeout setting** (`config.py:~86`): "The write runs off the response path, so this bounds a detached task rather than user-visible latency" — now inverted: the write is awaited in the request, so this setting is exactly what caps the latency a slow database can add to a turn. Worth rewording so the next person tuning it has the right picture. (Same vintage: the orphaned "Detached writes need a strong reference…" comment above `active_retention_days()` in `turn_log.py`.)

That's it — nothing else rises to blocking. Nice work landing this. 🙌

dannon:
Intentionally not fixing #1 -- we want to keep the error context. This is a public service and the beta logs get read by maintainers anyway, so a prompt fragment sitting in an exception is fine, and useful when working out why a write failed. Should have said so earlier rather than leaving it looking unaddressed.

#2 was real -- fixed in bc4b3910, both the timeout comment and the orphaned one above `active_retention_days`.

NoopDog (approved):
Approving — this is in great shape. Two tiny non-blocking notes that can ride the merge:

1. **README env-var comment** (`backend/api/README.md:~124`): with the coupled predicate, `ASSISTANT_TURN_LOG_PURGE_ENABLED=false` stops recording too, not just the sweep — worth the comment saying so ("false stops both the sweep and recording") so an operator doesn't pause deletion thinking the corpus keeps growing.

2. **PR description**: the "Notes for review" section still describes the write as running "off the response path as a detached task" — worth updating that paragraph to the inline design before merge, since the description becomes the merge record.

Thanks for the careful iteration on this one — the final shape (inline fail-open write + one retention predicate) is a really clean landing. 🚀

PR #1635, opened 2026-08-06
fix: fallback organism image for ga2 was gitignored

d-callan:
## Description

ga2 fallback images for when we dont have an image for an organism. this is what i had locally from some time ago, though we dont necessarily need to use this either if someone would like to do better.


frano-m (approved):
Approved thanks @d-callan

PR #1641, opened 2026-08-10
feat: ensure data output by initial catalog build step has consistent order (#1490)

### LINKED ISSUE #1490: Sort catalog outputs to avoid diff noise
Currently, some outputs of the catalog build (particularly the QC report) includes lists that are not sorted, leading messy diffs when the catalog is rebuilt and the items happen to have a different order. Find any places where lists are output from the catalog build and ensure that they're sorted at some point.

hunterckx:
## Description

Updates catalog build to be sorted or to retain input order, in order to keep the order consistent between runs:

- The inconsistent taxonomy IDs list is sorted
- Missing taxonomy tree ranks are derived by filtering rather than set difference (thus retaining order)
- The outbreak taxonomy ID list is sorted when read, giving a consistent order to anything derived from it
- The genomes dataframe is sorted before any QC checks that involve it
- The primary data dataframe is sorted before being saved
- The dbt tests are sorted

Various other outputs were already sorted or derived from something with a consistent order, including, I believe, all outputs from the TypeScript part of the build process

Both catalogs have been built, so diffs should be more meaningful immediately going forward

## Related Issue

Closes #1490


NoopDog (commented):
Reviewed — nice, well-targeted determinism work, and I appreciate that each sort carries a comment explaining *why* (set-iteration order, dbt thread-finish order). Verified the data files show no data loss (record counts preserved — assemblies/organisms 5506, GA2 assemblies 10696, `genomes-from-ncbi.tsv` 5507 with an identical accession set); the value churn is reordering plus a genuine upstream NCBI taxdump change (`data-build-meta.json` md5 `a5cf…` → `3bff…` on 2026-08-08), not structural change.

**Logic — all correct:**
- `report_inconsistent_taxonomy_ids` / `get_outbreak_taxonomy_ids` — `sorted(…, key=int)` over sets → total, deterministic.
- `do_taxonomy_tree_checks` — set-difference → order-retaining filter; same result set (levels are unique), `present_ranks` stays a set for O(1) lookup.
- `genomes_df` — moving `.sort_values("accession")` ahead of the QC checks is a strict improvement (those checks previously ran on nondeterministic merge order); `accession` is unique there, so the order is total.
- `transform.py` dbt results — stable sort on `test_name`. 👍

**One thing worth addressing (@hunterckx):**
`primarydata_df.sort_values("accession")` sorts on a **non-unique** key — `primarydata_df` is exploded to one row per SRA sample (`explode("sra_sample_acc")`), so many rows share an `accession`. pandas' default `sort_values` is `kind="quicksort"` (not stable), so rows sharing an accession aren't given a guaranteed relative order by the sort — their order still rides on upstream fetch/explode order rather than being pinned. So the "primary data sorted before saving" goal is only partially met for the multi-sample rows. Suggest a total key: `.sort_values(["accession", "sra_sample_acc"])` (or `kind="stable"` if upstream order is trusted). No crash risk (the column exists), and note this path is behind `extract_primary_data` (default `False`), so it wasn't exercised by the committed rebuild — worth a manual check when you next run it with primary data on.

**Minor:** the dbt test sort assumes `test_name` is unique; if two test nodes can share a name, ties stay in thread-finish order (add a secondary key like `status`/`message`). Fine if names are always unique.

**Observation (not a defect):** since the build fetches live NCBI/UCSC data, this committed rebuild mixes the ordering change with data drift, so the diffs can't be eyeballed as order-only — the committed data snapshots 2026-08-08 upstream state. That's inherent, and making future diffs meaningful is exactly the point of this PR.

Looks good to merge once the primary-data sort key is made total (or you're happy leaving it given the default-off path).

NoopDog (approved):
Approving — thanks for the quick turnaround. ✅

The `kind="stable"` fix (416181bf) resolves the main finding:
- `genomes_df` — `accession` is unique there, so no ties; stable-sort is a harmless defensive no-op.
- `primarydata_df` — this was the real one (non-unique key after `explode`). A stable sort now pins tied rows (same accession) to their input order, closing the "quicksort scrambles ties" failure mode.

Reconfirmed the data files are unchanged by the fix (build.py only), so the earlier verification still holds: no data loss, record counts preserved, the value churn is reordering + the genuine upstream taxdump change.

Two optional, non-blocking notes for the future (fine to leave):
- `kind="stable"` makes primary-data determinism depend on upstream (NCBI) response order; a fully total key `["accession", "sra_sample_acc"]` would pin the within-accession sample order too. The current approach is reasonable, especially as the path is behind `extract_primary_data` (default off).
- The dbt test sort keys on `test_name`; if two test nodes can ever share a name, add a secondary key (`status`/`message`) so ties don't fall back to thread-finish order.

Nice determinism win overall.

PR #1650, opened 2026-08-11
chore: add tracked .nvmrc and read it in workflows via node-version-file (#1648)

### LINKED ISSUE #1648: chore: add tracked .nvmrc and read it in GitHub workflows (node-version-file)
## Problem

The repo pins Node via `package.json` `engines.node` (22.13.0) and hard-codes `node-version: "22.13.0"` in the GitHub workflows, but has no `.nvmrc`. A tracked `.nvmrc` provides a single pin file that both local version managers (nvm/fnm/mise/`n auto`) and CI can read.

## Proposal

1. Add `.nvmrc` containing `22.13.0` (remove any `.gitignore` entry for it if present).
2. Replace hard-coded `node-version` in the workflows with:

```yaml
with:
  node-version-file: ".nvmrc"
```

This prevents the workflows from drifting when Node is bumped: a bump then only touches `.nvmrc` and `engines.node`, and CI follows automatically.

NoopDog:
Closes #1648

## What changed

- Added a tracked `.nvmrc` containing `22.13.0` and removed the `.gitignore` entry (plus its now-empty `# nvm` comment block) that was suppressing it.
- All five existing `actions/setup-node@v4` steps (`run-checks.yml` ×3, `check-pr-title.yml`, `publish-release.yml`) now use `node-version-file: ".nvmrc"` instead of a hard-coded `node-version: "22.13.0"`.
- Two jobs that run `npm ci` but had **no** setup-node at all — the scheduled catalog build in `update-catalog.yml` and the `api-tests` job in `run-checks.yml` — now get the same pinned setup-node step. These previously ran on whatever Node the runner shipped.
- The four hand-rolled `actions/cache@v4` (`~/.npm`) steps were replaced by setup-node's built-in `cache: "npm"` input — same lockfile-keyed caching, minus 16 lines, plus `restore-keys` fallback.
- The README's stale "Node.js version `22.12.0`" mention now points at `.nvmrc` instead of naming a version.

## Why

The Node pin lived in hard-coded workflow strings with no `.nvmrc` for local version managers (nvm/fnm/mise) to read, and the copies were already drifting (README said 22.12.0). With this change a Node bump touches `.nvmrc` + `engines.node`, and CI follows automatically.

## Assumptions I made

- The Dockerfile's `FROM node:22.13.0-alpine` stays a manual pin — base images can't consume `.nvmrc`. Considered and declined adding a lockstep comment or `ARG` plumbing; a future Node bump must touch the Dockerfile by hand.
- `engines.node` (and the `@types/node` pin) intentionally remain as duplicates of the version: `.nvmrc` was chosen over pointing `node-version-file` at `package.json` because local version managers only read `.nvmrc`. A stale `engines` self-reports via EBADENGINE warnings.
- Pinning the two previously-unpinned jobs (catalog cron, api-tests) is desired hardening, not scope creep — they build catalog artifacts that should be produced on the same Node CI validates.
- `publish-release.yml` gets the pin but no npm cache — it installs nothing.

## How to verify

Definition of done from #1648, mapped to checks:

- [x] **`.nvmrc` exists, tracked, containing `22.13.0`** — `cat .nvmrc` on this branch; `git check-ignore .nvmrc` exits 1 (not ignored).
- [x] **No hard-coded `node-version` remains in any workflow** — `grep -rn "node-version:" .github/workflows/` → zero matches; `grep -rn "node-version-file" .github/workflows/` → 7 sites.
- [x] **CI follows the pin automatically** — the checks on this PR run on Node 22.13.0 resolved from `.nvmrc` (see the setup-node step log line "Resolved .nvmrc as 22.13.0").
- [x] **Caching still works** — the setup-node step logs a cache hit for `~/.npm` keyed on `package-lock.json` on the second run.


frano-m:
Reviewed — nice clean-up, this is a solid improvement. 👍

## What's done well
- **Single source of truth done right** — `.nvmrc` read via `node-version-file: ".nvmrc"` in every `setup-node` step across all four workflows (`check-pr-title`, `update-catalog`, `run-checks` ×4, `publish-release`). Idiomatic and consistent.
- **Correctly tracks `.nvmrc`** — dropping it from `.gitignore` is exactly right so CI can read it.
- **README improved** — replaces a hardcoded (and *stale* — it said `22.12.0` while the real pin is `22.13.0`) version with a link to `.nvmrc`, so the docs self-correct on future bumps.
- **Net simplification** (−9 lines), removing the duplicated hardcoded versions from the workflows.

## One non-blocking suggestion
The centralization is partial — the Node version `22.13.0` is still hardcoded independently in:
- `Dockerfile:3` — `FROM node:22.13.0-alpine`
- `package.json` — `engines.node`

So the "one place to change it" goal isn't fully met: a future bump that touches only `.nvmrc` would leave **CI testing one Node version while the production Docker image builds/runs another** (silent drift), and a `package.json` `engines` mismatch would surface as `EBADENGINE` warnings (or failed installs under engine-strict).

Worth either aligning those two with `.nvmrc` (e.g. Docker `ARG NODE_VERSION` fed from a step that reads `.nvmrc`; `engines` kept in sync) or a short comment noting they're intentionally separate. Totally fine as a follow-up — didn't want it to silently drift later. (Minor aside: `@types/node` is also pinned to `22.13.0`, but that's a types-package version, a separate concern.)

frano-m (approved):
Approved!

PR #1662, opened 2026-08-13
fix: remove assemblyfavoritebutton from the ga2 assembly side column (#1661)

### LINKED ISSUE #1661: Remove AssemblyFavoriteButton from the GA2 assembly side column (favorites not wired up)
## Problem

The GA2 assembly detail side column renders `AssemblyFavoriteButton`, but GA2 has no favorites/auth wiring — so the button shows a permanent loading spinner.

Mechanism:

- `AssemblyFavoriteButton` renders a bare `<CircularProgress/>` while `!isConfigured || isAuthLoading` (`packages/shared/components/Favorites/components/AssemblyFavoriteButton/assemblyFavoriteButton.tsx`).
- `isConfigured` comes from the auth provider and is simply `loginEnabled` (`packages/shared/providers/authentication/provider.tsx`).
- GA2's site config never sets `loginEnabled` (implicitly false), so on GA2 `isConfigured` is always false and the button spins forever.

## Task

Remove the button from the GA2 site:

- `sites/ga2/views/EntityView/assembly/components/Side/ga2/side.tsx` — drop the `<AssemblyFavoriteButton accession={assembly.accession} />` render and its import.
- Check the surrounding layout still reads correctly without the action (heading/section spacing).

The shared component and BRC's usage stay as-is — favorites remain a BRC feature. If GA2 wires up auth/favorites later, the button can be reinstated then.

## Acceptance criteria

- No `AssemblyFavoriteButton` (or other Favorites imports) under `sites/ga2/`.
- GA2 assembly detail side column renders without a spinner/action; no visual regression otherwise.
- GA2 builds and e2e pass.

frano-m:
## Summary

Closes #1661

The GA2 assembly detail side column rendered `AssemblyFavoriteButton`, but GA2 has no favorites/auth wiring — `isConfigured` in the auth provider is just `loginEnabled`, which GA2's config never sets, so the button rendered a permanent `CircularProgress` spinner.

- Removes the `AssemblyFavoriteButton` render and import from `sites/ga2/.../Side/ga2/side.tsx`.
- Deletes the local `side.styles.ts` (`StyledSection`) — it existed solely to wrap the button.
- **Component-level fix (review follow-up)**: the shared `AssemblyFavoriteButton` conflated two states behind one spinner. Now `!isConfigured` (permanent — login not enabled on the site) renders `null`, while `isAuthLoading` (transient) keeps the spinner. This also fixes the same live spinner on **BRC**'s assembly detail pages, where `NEXT_PUBLIC_LOGIN_ENABLED` is unset on every deployed environment.
- BRC's `Side` gates the button's `StyledSection` wrapper on the same `isConfigured` signal, so a nulled button doesn't leave an empty padded strip.
- The shared component and BRC's usage remain wired for the day login is enabled — the sign-in affordance returns automatically.

## Verification

- No `Favorites` imports remain under `sites/ga2/`.
- `tsc --noEmit`, Prettier + ESLint pass (pre-commit hook).
- Full unit suite passes; BRC + GA2 builds compile; BRC e2e 75/75, GA2 e2e 8/8 pass.

### DIFF
--- a/packages/shared/components/Favorites/components/AssemblyFavoriteButton/assemblyFavoriteButton.tsx
+++ b/packages/shared/components/Favorites/components/AssemblyFavoriteButton/assemblyFavoriteButton.tsx
@@ -6,7 +6,9 @@ import { useAuth } from "@repo/shared/providers/authentication/provider";
 import { type JSX } from "react";
 import type { Props } from "./types";
 
-export function AssemblyFavoriteButton({ accession }: Props): JSX.Element {
+export function AssemblyFavoriteButton({
+  accession,
+}: Props): JSX.Element | null {
   const {
     isAuthenticated,
     isConfigured,
@@ -16,7 +18,11 @@ export function AssemblyFavoriteButton({ accession }: Props): JSX.Element {
   const { error, isFavorited, isLoading, isToggling, toggleFavorite } =
     useAssemblyFavorites();
 
-  if (!isConfigured || isAuthLoading) {
+  // Login is not enabled on this site — a permanent state, not a loading one —
+  // so favorites can't be offered and the button renders nothing.
+  if (!isConfigured) return null;
+
+  if (isAuthLoading) {
     return <CircularProgress size={20} />;
   }
 
--- a/sites/brc-analytics/views/EntityView/assembly/components/Side/brc/side.tsx
+++ b/sites/brc-analytics/views/EntityView/assembly/components/Side/brc/side.tsx
@@ -1,6 +1,7 @@
 import { buildOrganismDetails } from "@brc/viewModelBuilders/viewModelBuilders";
 import { BackPageContentSideColumn } from "@databiosphere/findable-ui/lib/components/Layout/components/BackPage/backPageView.styles";
 import { AssemblyFavoriteButton } from "@repo/shared/components/Favorites/components/AssemblyFavoriteButton/assemblyFavoriteButton";
+import { useAuth } from "@repo/shared/providers/authentication/provider";
 import {
   buildAssemblyDetails,
   buildAssemblyResources,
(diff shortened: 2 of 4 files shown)


NoopDog:
The GA2 removal is correct and verified clean (no dangling imports, the styles file's only consumer went with it) — but the same spinner is live on **BRC's** assembly detail page too, and this PR doesn't reach it. Worth folding into this PR since it's the same root cause:

**BRC evidence:** `isConfigured` is `loginEnabled`, and BRC's is `process.env.NEXT_PUBLIC_LOGIN_ENABLED === "true"` (`site-config/brc-analytics/local/config.ts:31`) — but none of the four BRC env files (`local`, `dev`, `prod`, `docker`) sets `NEXT_PUBLIC_LOGIN_ENABLED`. So `isConfigured` is `false` on every deployed BRC environment, and `sites/brc-analytics/.../Side/brc/side.tsx:27` renders the same permanent `CircularProgress` this PR removes from GA2. (Confirmed sighting on BRC.)

**Suggested fix, at the component level** — the button conflates two different states:

- `isAuthLoading` — transient → spinner is right
- `!isConfigured` — permanent → should render **nothing**

```tsx
if (!isConfigured) return null;
if (isAuthLoading) return <CircularProgress size={20} />;
```

That one-line split in `AssemblyFavoriteButton` fixes BRC today and any future site, keeps the button self-healing if BRC later flips `NEXT_PUBLIC_LOGIN_ENABLED=true`, and makes the GA2 removal here belt-and-braces rather than load-bearing. (Setting the env var instead would turn on the whole login UI — a product decision, not a spinner fix.)

frano-m:
@NoopDog folded in as suggested — 91afa33c:

1. **Component-level split in the shared `AssemblyFavoriteButton`**: `!isConfigured` (permanent — login not enabled on this site) → renders `null`; `isAuthLoading` (transient) → spinner. This fixes the live spinner on BRC's assembly detail pages (confirmed against https://brc-analytics.org/data/assemblies/GCA_001008285_1) and self-heals if a deployment flips `NEXT_PUBLIC_LOGIN_ENABLED=true`.
2. **One knock-on handled**: BRC's `Side` wrapped the button in a padded `StyledSection`, which would have rendered as an empty white strip once the button nulled. The section is now gated on the same `isConfigured` signal, so unconfigured BRC side columns start cleanly at Organism Details — and the section + sign-in affordance return automatically if login is ever enabled.

The GA2 removal stays as-is (belt-and-braces, per your framing — plus reinstating it would have reintroduced its own empty-section wrapper for a null child).

Verified: `tsc`, lint, prettier, full unit suite, BRC build + e2e 75/75.

NoopDog (approved):
Approving — the fold-in is exactly right: `!isConfigured → null` (permanent state) vs `isAuthLoading → spinner` (transient) at the component level, which fixes the live spinner on BRC's assembly pages and self-heals if login is ever enabled. Nice catch on the empty-`StyledSection` knock-on — gating it on the same `isConfigured` signal means the two gates can't drift apart harmfully. Verified BRC's `Side` is now the sole consumer, and keeping the GA2 removal is the right call. 👍

PR #1664, opened 2026-08-13
feat: friendly 'workflow not found' page for stale/incompatible workflow urls (#1487)

### LINKED ISSUE #1487: Friendly 'workflow not found' page for stale/incompatible workflow URLs
Follow-up from #1471 (PR #1484 review by @NoopDog).

Collapsing the per-`trsId` configure route to a `?trsId=` query param intentionally removes the route-level compatibility gate — any `?trsId=` now renders. Two consequences worth polishing:

## Inconsistent failure modes for stale/bad workflow URLs

- **Old path-style URL** (`.../analyze/workflows/<trsId>`) → hard **404** (no route matches).
- **New-style URL with an unknown/incompatible `trsId`** (`.../analyze/workflows?trsId=<garbage>`) → the base page renders, then `getWorkflow(trsId)` throws → caught by the app-wide `ErrorBoundary` → generic error UI.

Same "workflow doesn't exist / isn't compatible" situation, two different surfaces, neither of which tells the user what's wrong.

## Proposal: a friendly "workflow not found" page

Catch the not-found / incompatible case on the `analyze/workflows` route and render a friendly state — e.g. "That workflow isn't available for this assembly" with a link back to the compatible-workflows list — instead of the generic error boundary. This also gives the old-URL CDN redirect (see #1471 / the handoff fix) a sensible destination when a `trsId` is genuinely gone.

Note: the UI never links incompatible workflows (the assembly/organism detail pages only surface compatible ones), so this only affects hand-edited/stale/indexed URLs — hence a fast-follow, not a blocker.

frano-m:
## Summary

Closes #1487 

> **Stacked on #1663** — base is `fran/1488-configure-workflow-urlsearchparams`; GitHub will retarget to `main` when that merges. Only the last commit (`3022cc5f`) is this PR's diff.

A stale or unknown `?trsId=` on the analyze/workflows routes used to throw from `getWorkflow` and surface the generic error page ("Error / An error occurred processing your request" + raw store message). Now it renders a friendly, scoped state.

- **`findWorkflow(trsId)`** — non-throwing catalog lookup in the workflows service, mirroring `findOrganism`.
- **`WorkflowGate`** — the `EntityDataGate` idiom one level down: renders `children` when the TRS ID matches a catalog workflow, `fallback` otherwise. Rendered below `EntityDataGate`, so the store is always loaded before the lookup.
- **`WorkflowNotFound`** — friendly centered state reusing findable-ui's error-page layout primitives: amber status icon, "Workflow not found" heading, "The requested workflow isn't available for this assembly/organism. It may have been removed, or the link may be out of date.", and a "View Available Workflows" action.
- **Pages own the fallback** — all four analyze/workflows pages (BRC + GA2 × assembly/organism) pass `WorkflowNotFound` as the gate's `fallback` with a per-page back-link (assembly → compatible-workflows list, organism → organism detail). The views themselves are untouched.

Scope note: this handles the catalog-miss case (stale/hand-edited/indexed URLs). Compatibility gating for a real-but-unoffered workflow has no ready machinery in these views and the UI never links such workflows — deferred per the issue's own framing.

## Testing

- Unit: `findWorkflow` (hit/miss) and `WorkflowGate` (children/fallback) — 624 total pass.
- E2E: 2 new cases assert the not-found heading and back-link render for a garbage `trsId` on both the assembly and organism routes — BRC 81/81, GA2 8/8 pass.
- Both site builds compile; `tsc`, Prettier, ESLint clean.

### DIFF
--- a/packages/shared/components/workflow/WorkflowGate/types.ts
+++ b/packages/shared/components/workflow/WorkflowGate/types.ts
@@ -1,0 +1,7 @@
+import { type ReactNode } from "react";
+
+export interface Props {
+  children: ReactNode;
+  fallback: ReactNode;
+  trsId: string;
+}
--- a/packages/shared/components/workflow/WorkflowGate/workflowGate.tsx
+++ b/packages/shared/components/workflow/WorkflowGate/workflowGate.tsx
@@ -1,0 +1,23 @@
+import { findWorkflow } from "@repo/shared/services/workflows/entities";
+import { type JSX } from "react";
+import type { Props } from "./types";
+
+/**
+ * Gate content on the TRS ID matching a catalog workflow. Renders `children`
+ * for a known workflow and `fallback` otherwise, letting the page decide how
+ * a stale or unknown workflow URL is surfaced. Must be rendered below
+ * EntityDataGate, which guarantees the workflows cache is loaded before the
+ * lookup runs.
+ * @param props - Component props.
+ * @param props.children - Content to render when the workflow exists.
+ * @param props.fallback - Content to render for an unknown TRS ID.
+ * @param props.trsId - Workflow TRS ID.
+ * @returns Children when the workflow exists, fallback otherwise.
+ */
+export function WorkflowGate({
+  children,
+  fallback,
+  trsId,
+}: Props): JSX.Element {
+  return <>{findWorkflow(trsId) ? children : fallback}</>;
+}
--- a/packages/shared/components/workflow/WorkflowNotFound/types.ts
+++ b/packages/shared/components/workflow/WorkflowNotFound/types.ts
@@ -1,0 +1,4 @@
+export interface Props {
+  entityContext: string;
+  href: string;
+}
--- a/packages/shared/components/workflow/WorkflowNotFound/workflowNotFound.tsx
+++ b/packages/shared/components/workflow/WorkflowNotFound/workflowNotFound.tsx
@@ -1,0 +1,62 @@
+import { ButtonPrimary } from "@databiosphere/findable-ui/lib/components/common/Button/components/ButtonPrimary/buttonPrimary";
+import { AlertIcon } from "@databiosphere/findable-ui/lib/components/common/CustomIcon/components/AlertIcon/alertIcon";
+import { SectionActions } from "@databiosphere/findable-ui/lib/components/common/Section/section.styles";
+import {
+  PRIORITY,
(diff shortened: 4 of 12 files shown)


NoopDog (approved):
Approving — verified the load-bearing pieces:

- **Gate ordering is sound**: `WorkflowGate` sits below `EntityDataGate` at all four call sites, so the synchronous `findWorkflow` lookup never runs before the store is loaded (and thanks to #1657, a *failed* load surfaces at the `EntityDataGate` level, so the gate never consults an empty-because-errored store).
- **The lookup can't miss on formatting**: #1663 guarantees every emitted `trsId` URL param is `formatTrsId`-normalized, which is exactly how the store is keyed — nice case of the stack ordering doing real work.
- **Back-link targets check out**: assembly → its compatible-workflows listing; organism → the organism detail page, which is the correct destination (it hosts the workflows section; no organism-scoped list route exists).
- The no-trsId organism path is a strict improvement too (friendly state instead of the view throwing into the generic error page), and the a11y shape is right (`h1` + link role, both pinned by the e2e assertions).
- Agree with the scope cut on compatibility gating — and the chosen copy happens to read correctly for the deferred case as well, so it won't need rewording later.

Two non-blocking notes:

1. The four pages repeat the ~10-line gate+fallback wiring with only `entityContext`/route varying — deliberate and fine at four sites; if a fifth appears, that's the moment for a small `buildWorkflowNotFound(entityContext, href)` helper.
2. The e2e cases assert the heading and link visibility but not the link's `href` — one `toHaveAttribute("href", …)` on the assembly case would pin the per-page back-link, which is the only logic the four pages actually differ in.

(To merge after #1663, its declared base.)

frano-m:
@NoopDog re: [your review](https://github.com/galaxyproject/brc-analytics/pull/1664#pullrequestreview-4923388609) —

1. **Four-site repetition** — agreed, leaving as-is; a fifth call site is the trigger for a `buildWorkflowNotFound` helper.
2. **Back-link `href` pinning — done in 1a693c80**, on both stale-URL cases (assembly → `ANALYZE_WORKFLOWS`, organism → `ORGANISM` detail), since the back-link target is the one thing the pages differ in. Verified live on chromium (2/2).

Also rebased the stack: this branch now sits on the current #1663 tip, which itself is rebased onto today's `main` (includes the #1662 merge).

PR #1673, opened 2026-08-14
feat: italicize scientific names (#632)

### LINKED ISSUE #632: italicize scientific names?
i first noticed this on the priority pathogen details page, but now i see it i cant stop seeing it all over the place.

frano-m:
## Summary

Closes #632

Implements Phases 1 + 2 of the plan on the issue: a shared `ScientificName` component (semantic `<i>`, JSDoc'd with the typographic rule — genus-and-below italic; strains, serotypes, isolate codes and above-genus ranks stay roman) threaded through every markup-capable render of a species name on both sites.

**Phase 1 — builders (both sites):**
- Species columns in all four entity lists (BRC `buildGenomeSpecies` / `buildOrganismTaxonomicLevelSpecies`, GA2 `buildAssemblySpecies` / `buildOrganismSpecies`). The organism-detail nested assembly tables override the cell's primary label with the accession *after* the spread, so accessions stay roman automatically.
- Both sites' organism-details side sections (shared `buildOrganismDetails` species link).
- Priority-pathogen taxon links — **rank-gated** on the catalog's `taxonNameField` discriminator (genus/species italicize; family/order names like `Arenaviridae`/`Mucorales` and descriptors like "species complex" stay roman).

**Phase 2 — stepper/workflow surfaces:**
- ENA `CollectionSelector` `scientific_name` and `AssemblySelector` species columns via a shared, null-safe `renderScientificName` cell renderer (`Table/components/TableCell/components/ScientificName/utils.tsx`).
- ENA `DataSelector` species label and `AssemblySummary` species cell.
- **Both sites' Workflows tables** via `renderWorkflowSpecies` (shared `viewModelBuilders`), which keeps the `Any` applicability sentinel roman — `Any` promoted from 10 scattered literals to the `TAXON_ANY` constant.

**Deliberate decisions (documented in code):**
- **Hero titles stay roman** (commented in both `buildOrganismHero`s): at H1 scale the italic reads as styling, not nomenclature.
- Catalog species fields wrap wholesale; embedded `sp.`/`cf.` qualifiers inheriting the italic is an accepted trade-off (stated in the component JSDoc).
- `priorityPathogen.name` (mixed common/scientific) untouched pending the catalog-flag decision; Phase 3 (sunburst, rank-aware SVG labels) and string-typed surfaces (breadcrumbs, titles, warning strings) deferred per the issue.

## Testing

- Updated `species.test` asserts the wrapped labels; full unit suite passes (622).
- `tsc --noEmit`, ESLint, Prettier pass; both site builds compile; BRC e2e 81/81 and GA2 8/8 pass on the final tree.
- Reviewed via high-effort multi-agent code review; all 10 findings addressed or explicitly traded off (see commits).


frano-m:
The three suppressed notes on the last Copilot review are addressed in the latest commit:

1. **Orphaned `columnDef.tsx` — the significant one.** Both `columnDef.ts` and `columnDef.tsx` existed in the AssemblySelector table folder (my rename-back left the stale inline-lambda `.tsx` behind), and since the resolver prefers `.tsx`, the stale copy was the one actually bundling while the named-renderer `.ts` was dead code. Behavior was identical either way (both italicize the species), but the duplicate is now deleted and `columnDef.ts` with `renderScientificName` is the single source — verified by a clean build.
2. **`DataSelector` species can be undefined** — the `ScientificName` wrapper now renders conditionally, so no empty `<i>` when the workflow-entity context is absent.
3. **JSDoc drift** — `ITALIC_TAXON_NAME_FIELDS`' comment now matches the allowlist exactly: genus and species only; family/order names, strains and descriptors stay roman.

`tsc`, lint, prettier and the full unit suite pass; BRC build compiles clean. (Local e2e skipped — port 3000 in use — CI covers it.)

frano-m:
Latest suppressed note addressed in 092677c6 — the three remaining `taxonomyId: "Any"` literals on the injected workflows (DEA, Logan Search, LexicMap) now use `TAXON_ANY`, along with the `commonNames` fallback array. All "Any" sentinel values in `WorkflowsView/utils.ts` now flow from the one constant; the remaining "Any" strings in that file are JSDoc prose only.

packages/shared/views/EntityView/components/ConfigureWorkflowInputs/components/Main/components/Stepper/components/Step/ReferenceAssemblyStep/components/AssemblyData/components/AssemblySummary/assemblySummary.tsx
```diff
@@ -54,7 +55,11 @@ export const AssemblySummary = ({
               {getSelectedRowModel().rows.map((row) => (
                 <TableRow key={row.id}>
                   <TableCell>{row.original.accession}</TableCell>
-                  <TableCell>{row.original.taxonomicLevelSpecies}</TableCell>
+                  <TableCell>
+                    <ScientificName>
```
NoopDog: Non-blocking consistency nit: this wraps `taxonomicLevelSpecies` without a null-guard, whereas `DataSelector` guards (`taxonomicLevelSpecies && …`) and the shared `renderScientificName` guards by design (its JSDoc notes it passes absent values through so a missing name doesn't render an empty italic element). If this field were ever empty/undefined you'd get an empty `<i></i>`. It's effectively always populated for a selected assembly, so this is low-risk — just calling out the inconsistency with the guarding used elsewhere.

packages/shared/viewModelBuilders/viewModelBuilders.tsx
```diff
@@ -635,3 +638,17 @@ export function getGenomeStrainText(
...
+ * Renders a workflows-table species cell as a scientific name, keeping the
+ * "Any" applicability sentinel roman.
+ * @param ctx - Cell context.
+ * @returns Species cell content.
+ */
+export function renderWorkflowSpecies<T>(
+  ctx: CellContext<T, unknown>
+): ReactNode {
+  const value = ctx.getValue<string | undefined>();
+  if (!value || value === TAXON_ANY) return value ?? null;
```
NoopDog, line 652: Non-blocking test-coverage note: the new `species.test` covers the happy wrapping well, but the branches with actual conditional logic aren't directly tested — this `Any`-stays-roman short-circuit (a headline feature), `renderScientificName`'s null passthrough, and the priority-pathogen rank-gating (`ITALIC_TAXON_NAME_FIELDS`: genus/species italic vs family/order roman). A couple of small unit tests would lock in the behaviors most likely to regress silently if this code is refactored later.

NoopDog (approved):
LGTM — clean, well-architected, and presentation-only. I checked the main risk (sort/filter/search): every table column keeps its `accessorKey`/`filterFn`/`id` and only adds a `cell` renderer, and the list builders already returned an object (never sorted on `species.label`), so wrapping the label in `<i>` is functionally inert for faceting/sorting. `<i>` is the spec-correct element for taxon names, the `Any` sentinel and rank-gating are handled correctly, GA2 has parity, and I ran `species.test.tsx` locally (6/6 green). Two non-blocking nits inline.

packages/shared/views/EntityView/components/ConfigureWorkflowInputs/components/Main/components/Stepper/components/Step/ReferenceAssemblyStep/components/AssemblyData/components/AssemblySummary/assemblySummary.tsx
frano-m: Done in 82506fd7 — the `AssemblySummary` species cell now guards like `DataSelector` (`taxonomicLevelSpecies && …`), so an absent value renders nothing instead of an empty italic element. Consistent with `renderScientificName`'s documented passthrough.

packages/shared/viewModelBuilders/viewModelBuilders.tsx
frano-m: Done in 82506fd7 — added `tests/viewModelBuilders/scientificName.test.tsx` covering exactly the conditional branches: `renderScientificName` (wrap + null passthrough), `renderWorkflowSpecies` (wrap + `Any`-stays-roman + null passthrough), and the priority-pathogen rank gating (species and genus italicize; family stays roman) asserted through `buildPriorityPathogenDetails`' link children. 8 new tests, suite at 630.

frano-m:
Latest suppressed note addressed and branch rebased:

- **Absent `taxonName` guard** — the rank-gated wrap now checks the name is present first, so a missing `taxonName` passes through unchanged (previous behavior) instead of rendering an empty italic element. Same guard convention as `DataSelector`, `AssemblySummary` and `renderScientificName`.
- **Rebased onto current `main`** (includes the #1677 merge) — all 7 commits replayed cleanly, `tsc` + 630 unit tests green post-rebase.

frano-m:
Latest pair of suppressed notes addressed in b0128f44 — the sentinel test now imports `TAXON_ANY` and uses it for both the input and the expected passthrough, so the test tracks the sentinel's source of truth. (Worth noting the flip side consciously: a literal would have pinned the user-visible string and broken loudly on an intentional sentinel change — but since the sentinel flows from the same constant into both the data builders and the renderer comparison, the constant keeps the test aligned with the actual contract.)

PR #1679, opened 2026-08-16
feat: validate source data during build by reading into pydantic models (#1547)

### LINKED ISSUE #1547: Validate catalog source data when read via Python
The source YAML files for the catalog are validated statically by the pre-commit hook and GitHub checks workflow, but not at runtime. For extra safety, we can use the Pydantic models to validate the YAML data when it's read in `build.py`.

hunterckx:
## Description

Updates `build.py` to use LinkML's `YAMLLoader` with the generated Pydantic models to load catalog source data. In addition to validating that the source data matches the schema, this also ensures that optional fields are always represented by a column in the returned dataframe (even if they're never used in the source data).

I've run the catalog builds to ensure they behave the same but have not included the results here since this involves no changes to the output shape.

## Related Issue

Closes #1547

### DIFF
--- a/catalog/py_package/catalog_build/build.py
+++ b/catalog/py_package/catalog_build/build.py
@@ -13,10 +13,11 @@ import bs4
 import duckdb
 import pandas as pd
 import requests
-import yaml
 from bs4 import BeautifulSoup
+from linkml_runtime.loaders import YAMLLoader
 from requests.exceptions import ConnectTimeout
 
+from .generated_schema import schema
 from .load import do_dlt_load
 from .qc_utils import (
     format_list_section,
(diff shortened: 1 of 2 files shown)


NoopDog (commented):
Nice change — validating source against the schema at read time is a good move. I ran the old and new read paths against the real `catalog/source/*.yml` and confirmed values are preserved: enums come out as plain strings (thanks to `use_enum_values=True` on the generated base model), and all three models exactly cover the current source fields so `extra="forbid"` doesn't reject anything. Three small notes below; none blocking.

catalog/py_package/catalog_build/build.py
```diff
@@ -205,25 +206,36 @@ def post_ncbi_request(url, json_data, batch_size=1000, min_batch_size=50):
...
+
+    Args:
+        yaml_path: Path of the YAML file to read.
+        schema_model: Pydantic model representing the root schema class for the YAML file.
+        list_key: Key of the root object in which the list of entities is held.
+
+    Returns:
+        df: Dataframe representing the list of entities.
+    """
+    yaml_data = YAMLLoader().load(source=yaml_path, target_class=schema_model)
```
NoopDog, line 221: **1. Design question — why `YAMLLoader` here?** `iwc_manifest_to_workflows_yaml.py:67` already validates through these same generated models with `Workflows.model_validate(yaml.safe_load(fh))`. The equivalent here — `schema_model.model_validate(yaml.safe_load(open(yaml_path)))` then `getattr(..., list_key)` — would validate identically, keep `yaml.safe_load` (no `yaml` import removal), stay consistent with that existing file, and avoid pulling `linkml-runtime` into `build.py`'s import path. Not wrong as written; just curious whether the LinkML loader was chosen deliberately (e.g. for a coercion it does that `model_validate` wouldn't), since this introduces a second mechanism for the same job.

catalog/py_package/catalog_build/build.py
```diff
@@ -205,25 +206,36 @@ def post_ncbi_request(url, json_data, batch_size=1000, min_batch_size=50):
...
 def read_organisms(organisms_path):
-    with open(organisms_path) as stream:
-        organisms_data = yaml.safe_load(stream)["organisms"]
-        # Convert taxonomy_id to string to ensure consistent type handling
-        for organism in organisms_data:
-            organism["taxonomy_id"] = str(organism["taxonomy_id"])
-        return pd.DataFrame(organisms_data)
+    return read_dataframe_from_yaml(
+        organisms_path, schema.Organisms, "organisms"
+    ).astype({"taxonomy_id": "string"})
```
NoopDog, line 232: **2. `taxonomy_id` dtype change worth a sanity check.** The old path produced an `object` column of Python `str`; `model_dump()` yields `int`, and `.astype({"taxonomy_id": "string"})` gives a pandas nullable `StringDtype` column. The values are identical (I diffed them), but the dtype differs (object → StringDtype). Could you confirm the downstream dlt load / dbt treats these the same? Likely fine, and probably already covered by your build-output check — flagging because it's the one dtype that actually changed.

catalog/py_package/catalog_build/build.py
```diff
@@ -205,25 +206,36 @@ def post_ncbi_request(url, json_data, batch_size=1000, min_batch_size=50):
...
 
 
 def read_outbreaks(outbreaks_path):
     if outbreaks_path is None:
         return None
-    with open(outbreaks_path) as stream:
-        return pd.DataFrame(yaml.safe_load(stream)["outbreaks"])
+    return read_dataframe_from_yaml(outbreaks_path, schema.Outbreaks, "outbreaks")
```
NoopDog, line 238: **3. Absent optionals now come through as `None` rather than `NaN`.** With `model_dump()`, an outbreak lacking `highlight_descendant_taxonomy_ids` (11 of 24 rows) yields `None` instead of the `NaN` the old `pd.DataFrame(...)` produced. This is the intended fix — the column is now always present — just worth confirming dlt infers the same column type from `None`-filled cells as it did from `NaN`.

catalog/py_package/catalog_build/build.py
hunterckx: I figure since we're using LinkML for the schema it makes sense to use LinkML to load the data, and this doesn't increase the number of imports or increase the number of dependencies (if you take into account indirect ones)

catalog/py_package/catalog_build/build.py
hunterckx: Double-checked and I don't see anywhere that the adjusted type would make a difference. dlt/dbt in particular aren't actually relevant because the column gets converted back to integers before being loaded into the database

catalog/py_package/catalog_build/build.py
hunterckx: Not an active concern since currently only the primary taxonomy ID columns are loaded into the database, but a good thing to keep in mind in the future at least. (Also worth noting that this PR doesn't make things any more fragile than they were, and in fact having the column present but with an unhelpful type is probably less problematic than having it absent entirely like it would have been previously)

NoopDog (approved):

Issue #1477, opened 2026-07-21
Surface an error when the entity-cache fetch fails (data pages render blank)


frano-m:
## Background

The data explore list pages (`data/assemblies`, `data/organisms`) — and the other data pages generally — render their content only once the client-side entity cache has loaded, gated via `EntityDataGate` / `useEntitiesLoaded()`. The cache is populated by `useEntities()` in `_app`, which fetches the runtime `/api/*.json` catalogs.

## Problem

If that fetch fails (e.g. a 404 on a misconfigured deploy, a network error, or malformed JSON), `useEntities()` never flips `isLoaded` to true, so the gate renders its `null` fallback indefinitely. The result is a permanently blank content area with **no error surfaced and no retry** — the user can't tell the difference between "still loading" and "failed".

There is also no loading indicator during the fetch (the gate's default fallback is `null`), so a slow load shows a blank area rather than a spinner/skeleton.

## Proposal

- Surface an error when the entity-cache fetch fails (throw into the existing `ErrorBoundary` in `_app`, or expose an error state from `useEntities()` that the gate can render), so a failed load shows the error UI instead of a blank page.
- Optionally add a loading indicator (skeleton/spinner) while the cache loads, weighing it against the current intentional no-flash-when-cached behaviour.

## Notes

This is app-wide behaviour, not specific to any one page — it affects every `EntityDataGate` consumer. Surfaced during the code review of #1472 (stop inlining the catalog into the explore list pages), which made the store-load dependency the sole source for those tables.

Issue #1472, opened 2026-07-21
Stop inlining the full catalog into data/assemblies.html and data/organisms.html (~15 MB duplicated)


NoopDog:
## Problem

The two explore/list pages ship the **entire catalog inlined into the HTML**, and the same data is *also* fetched client-side — so the dataset is downloaded and parsed twice.

| Page | HTML size | Rows inlined in `__NEXT_DATA__` |
|---|---|---|
| `out/data/organisms.html` | **8.2 MB** | 1,975 organisms |
| `out/data/assemblies.html` | **6.9 MB** (7.2 MB `__NEXT_DATA__`) | 5,499 assemblies |

Every other page in the app is 2–3 KB.

### Where it comes from

`pages/data/[entityListType]/index.tsx` → `getStaticProps` sets `props.data = await fetchAllEntities(...)` for `EXPLORE_MODE.CS_FETCH_CS_FILTERING`. Next serializes that whole response into the page's `__NEXT_DATA__`.

### Why it's redundant

The client separately loads the same catalog from the served API files (`app/services/workflows/loader.ts` → `loadEntities` → `fetch("/api/assemblies.json")`, backed by `out/api/assemblies.json` = 8.8 MB and `out/api/organisms.json` = 12 MB). And the page body renders nothing at build regardless (0 visible chars — content is client-hydrated). So the inlined `pageProps.data` is a second, eagerly-parsed copy of data the client fetches anyway.

### Cost

- The browser must download the ~7–8 MB HTML **and** parse the inlined JSON before the page is interactive, then fetch the ~9–12 MB API file on top — doubling transfer and adding a large main-thread JSON parse.
- Bloats the static export output and every CDN cache entry for these two routes.

## Proposed fix

1. **Verify the consumer**: confirm whether findable-ui's `ExploreView` actually reads `props.data` on the client, or whether the explorer re-derives its rows from a client-side source (the findable-ui `database` seeded via `seedDatabase`, and/or the `/api/*.json` fetch). The `CS_FETCH_CS_FILTERING` mode name ("client-side fetch") suggests the explorer owns its own data load.
2. **If redundant** (expected): drop `props.data` from `getStaticProps` for these list pages so the HTML falls from ~7–8 MB to a few KB. The client already has a data path.
3. **If the explorer needs initial data**: use a lighter seed (e.g. only the first page of rows) rather than the full catalog, or rely on the `/api/*.json` fetch that already happens.

## Impact

[...]

Issue #1341, opened 2026-06-10
Pangenome: populate pangenomes.yml from workflow output and build pangenomes.json


NoopDog:
Parent epic: #1322

Populate the pangenome instance and produce an app-consumable export:

- **Populate `catalog/source/pangenomes.yml`** (the `plasmodium-vivax-v1` entry) from the **Galaxy workflow output** (#1339) — version, create date, anchor, member assemblies, and the BRC Data Store output-file links.
- **Build `catalog/output/pangenomes.json`** consumable by the app, following the existing entity build/export pattern (as for `assemblies.json` / `organisms.json`).
- **Retrievable by species taxonomy ID** — the organism page for a taxon (e.g. 5855) can fetch that species' pangenome(s).
- **Validations, like other entity types** (in `validate_catalog.py`): schema validation plus the cross-catalog check below.

## Member schema (v0)

`member_assemblies` is a **map keyed by assembly accession** — the accession is the UCSC hub `genome=` arg used for the #1337 browser links. Each member's value:

- `strain` — *optional* display name (e.g. `PvP01`). Not every member will have one.
- `is_anchor` — *optional*, defaults to `false`. Exactly **one** member must set `is_anchor: true` (the reference the alignments are anchored on); the build fails if zero or more than one member is flagged.
- `catalog_accession` — *optional*; only needed when the hub accession differs from the in-catalog accession. Defaults to the map key. Sal-I is the one v0 case: the hub carries the **GCA** (`GCA_000002415.2`) but the catalog holds the **GCF** (`GCF_000002415.2`), so its member sets `catalog_accession: GCF_000002415.2`.

### Cross-catalog validation

Every member's **catalog accession** (its `catalog_accession`, or the map key when absent) must exist in the BRC assemblies catalog — **exact accession match, no GCA↔GCF aliasing**. The Sal-I GCA→GCF exception is an explicit per-member field, not a derived/fuzzy mapping. Validation fails if any member's catalog accession isn't present.

### Example

```yaml
# catalog/source/pangenomes.yml
- id: plasmodium-vivax-v1
  species_taxonomy_id: 5855
  version: "2026-05"
  member_assemblies:
    GCA_900093555.2:                       # PvP01
      strain: PvP01
      is_anchor: true
    GCA_000002415.2:                       # Sal-I
      strain: Sal-I
      catalog_accession: GCF_000002415.2   # hub uses the GCA; catalog holds the GCF
    GCA_914969965.1:
[...]

PR #1680, opened 2026-08-17
feat: prerender real content for the organism detail pages (#1473)

### LINKED ISSUE #1473: Prerender real content for entity detail pages (organism/assembly) instead of empty client-hydrated shells

frano-m:
## Summary

First page-type slice of #1473 (organism detail, both sites) — per the design recorded on the issue and the Slack sync: prerender the full page from build-available data, one page type at a time. **Does not close #1473** — assembly detail and the two analyze pages follow the same recipe in follow-up PRs.

### Before / after

The built organism pages' `<body>` had **0 visible characters** (everything client-rendered behind `EntityDataGate`). They now prerender the hero (species, taxonomy/priority tags), tabs, the organism-specific workflow cards, and the assemblies table headers — measured on real build output.

### How

- **`makeEntityStaticProps` gains an optional `augmentData` hook** (`<R, D extends R>`), letting a page attach build-computed fields to the entity record it already ships.
- **Shared `loadOrganismWorkflowCategories(config, organism)`**: reads the site's `workflows.json` (same source the runtime store fetches; memoized per build worker, memo evicted on rejection like `seedDatabase`), and computes the organism's compatible categories via the existing pure `buildOrganismWorkflows`.
- **Feature-flag semantics preserved**: flags are per-user runtime state (`useFeatureFlag` defaults false; `setFeatureFlags` only allowlists URL toggles), so the build embeds the **flag-inclusive superset** and `WorkflowsSection` applies the runtime gate at render. `useFeatureFlag` returns false on the server and during hydration, so SSR markup and the hydration pass agree structurally. Verified on build output: the assembly-category workflow is in `__NEXT_DATA__` but not in SSR markup.
- **BRC pangenome** embedded per page from `catalog/output/pangenomes.json` (absent file reads as empty with a warning; any other failure — including a non-array payload — fails the build loudly). `useShowPangenome`/`Tabs`/`PangenomeSection` take the bundle via props.
- **`EntityDataGate` removed from the organism detail pages only** — everything they render is now props-driven.
- **Dead client pipeline removed**: with no store reader left, the per-visitor `/api/pangenomes.json` fetch, BRC pangenome loader/getter/routes module and their tests are deleted.

### Measurements (all 1,975 BRC organism pages)

- `__NEXT_DATA__`: median 2.3KB → 3.6KB (the workflow slice); max unchanged at 663.7KB (the record's own `genomes` — pre-existing). No catalog in any page; #1472 intact.
- Hydration: zero React errors on samples including the 489-assembly worst case; tables fill instantly from embedded props (no fetch wait, no flash).

### Judgment calls for review

1. **Virtualized table rows stay client-rendered** — findable-ui's `useVirtualization` needs a measured scroll container, so SSR emits headers only; rows hydrate immediately from props. Prerendering them means findable-ui-level work for the least SEO-valuable text (assembly accessions get their own prerendered pages in the next slice).
2. **The #1477 error page no longer surfaces on this route** — the page no longer consumes the client store, so a failed cache load surfaces on the next store-dependent route instead. Deliberate consequence of un-gating; flagging for confirmation.
3. **Follow-up noted on the issue**: embedded workflows carry full parameter specs (~4.1KB vs ~1.8KB rendered) — a render-shaped DTO would halve the slice but touches shared component typing.

## Testing

- `tsc --noEmit`, ESLint, Prettier, 627 unit tests pass; both site builds compile.
- Emitted-HTML content crawl + flag-semantics check + hydration console check on served build output (samples incl. the largest organism).
- Local e2e blocked by a port collision with a dev server; CI's run is the e2e signal.
- Reviewed via high-effort multi-agent code review: 10 verified findings — 8 fixed (headlined by the feature-flag tautology), 2 deferred with rationale (items 2–3 above).


packages/shared/services/staticGeneration/workflows/utils.ts
```diff
@@ -0,0 +1,65 @@
...
+ * @returns Workflow categories.
+ */
+export function loadWorkflowCategories(
+  staticLoadFile: string
+): Promise<WorkflowCategory[]> {
+  let promise = categoriesByFile.get(staticLoadFile);
+  if (!promise) {
+    promise = fsp
+      .readFile(staticLoadFile, "utf8")
+      .then((text) => JSON.parse(text) as WorkflowCategory[]);
```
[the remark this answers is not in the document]
frano-m: Done in 465a7979 — the workflows reader now validates the parsed payload is an array and fails with a clear shape error naming the file, matching the pangenome reader's guard.

sites/brc-analytics/services/staticGeneration/organism/utils.ts
```diff
@@ -0,0 +1,66 @@
...
+export async function augmentOrganismDetail(
+  organism: BRCDataCatalogOrganism
+): Promise<BRCOrganismDetail> {
+  const workflowCategories = await loadOrganismWorkflowCategories(
+    config,
+    organism
+  );
+  const pangenome = (await loadPangenomes()).find(
+    (p) => p.speciesTaxonomyId === organism.ncbiTaxonomyId
+  );
```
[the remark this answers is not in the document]
frano-m: Done in 465a7979 — the pangenomes catalog now memoizes as a `Map` keyed by `speciesTaxonomyId` (mirroring the former runtime store's keying), so the per-page lookup is O(1) regardless of how many pangenome bundles the catalog grows.

frano-m:
The suppressed note on the last Copilot review is addressed in the latest commit — `makeEntityStaticProps` is now overloaded so an augmented data type `D` can only be claimed when the `augmentData` that produces it is supplied; calling with two type arguments and no augmenter now fails compile instead of shipping `data as D` with undefined fields.

packages/shared/services/staticGeneration/workflows/utils.ts
```diff
@@ -0,0 +1,68 @@
...
+ * @returns Organism-compatible workflow categories.
+ */
+export async function loadOrganismWorkflowCategories(
+  config: () => Pick<SiteConfig, "entities">,
+  organism: OrganismContract
+): Promise<WorkflowCategory[]> {
+  const categories = await loadWorkflowCategories(
+    getWorkflowsStaticLoadFile(config)
+  );
+  return buildOrganismWorkflows(organism, categories, true);
```
NoopDog: **medium** — The flag-to-category mapping now lives in two places. This call hardcodes `true`, so the build emits the superset for *any* flag-gated category, but `workflowsSection.tsx` re-gates only `ASSEMBLY`.

Concrete failure: add a second flagged category to `buildOrganismWorkflows` and organism detail pages will render it to every user with the flag off — no type error and no test failure to catch it, because the build side deliberately ignores flags and the render side only knows about one category.

Consider having the augmenter emit which categories are flag-gated, or deriving both sides from a single constant, so `buildOrganismWorkflows` stays the only owner of that rule.

packages/shared/views/OrganismView/components/Main/components/WorkflowsSection/workflowsSection.tsx
```diff
@@ -1,32 +1,35 @@
...
- * @param props.organism - Organism.
+ * @param props.workflowCategories - Organism-compatible workflow categories.
  * @returns The workflows section.
  */
 export const WorkflowsSection = ({
   entityId,
-  organism,
+  workflowCategories: allWorkflowCategories,
 }: Props): JSX.Element => {
   const isAssemblyWorkflowsEnabled = useFeatureFlag("assembly-workflows");
```
NoopDog: **medium** — Other half of the duplicated gate (see the comment on `staticGeneration/workflows/utils.ts:25`). The filter just below silently re-implements a rule that `buildOrganismWorkflows` owns: "the `ASSEMBLY` category is behind the `assembly-workflows` flag".

It is correct today — `buildOrganismWorkflows`'s only flag effect is skipping that one category, so build-with-`true` plus this filter is exactly build-with-the-flag. The risk is drift: the two copies have no shared source, so a change to the gating rule in one place will not show up in the other.

sites/brc-analytics/services/staticGeneration/organism/utils.ts
```diff
@@ -0,0 +1,66 @@
...
+function loadPangenomes(): Promise<Map<string, Pangenome>> {
+  pangenomesPromise ??= fsp
+    .readFile(PANGENOMES_STATIC_LOAD_FILE, "utf8")
+    .then((text) => {
+      const pangenomes = JSON.parse(text) as Pangenome[];
+      if (!Array.isArray(pangenomes))
+        throw new Error("Pangenomes catalog is not an array");
+      return new Map(pangenomes.map((p) => [p.speciesTaxonomyId, p]));
+    })
+    .catch((error) => {
```
NoopDog: **low** — Net loss of test coverage. The deleted `loadPangenomes` tests covered missing-file, non-array payload, and store keying. The replacement logic here (ENOENT vs. rethrow, the non-array guard, memo eviction, keying by `speciesTaxonomyId`) has no tests at all.

Related: `loadWorkflowCategories` in `packages/shared/services/staticGeneration/workflows/utils.ts` is exported but never imported anywhere — presumably for tests that were not written. These are pure functions over a file path, so they are cheap to cover directly.

sites/brc-analytics/services/staticGeneration/organism/utils.ts
```diff
@@ -0,0 +1,66 @@
+import { type BRCDataCatalogOrganism } from "@brc/apis/organism";
+import { type Pangenome } from "@brc/apis/pangenome";
+import { config } from "@brc/config/config";
+import { loadOrganismWorkflowCategories } from "@repo/shared/services/staticGeneration/workflows/utils";
+import { promises as fsp } from "fs";
+import { type BRCOrganismDetail } from "./types";
+
+// The catalog build's pangenomes output; the sync-api script copies it to the
```
NoopDog: **low** — This coupling is documented in the comment but unenforced, and both sides fail soft.

Concrete failure: rename the catalog output and `sync-api-brc-analytics.sh` prints "Skipping optional pangenomes.json", this reader warns once per build worker, and all 1,975 pages ship with no Pangenome tab and no Pangenome section. The build is green and the feature is silently gone.

The pre-existing runtime loader had the same soft-fail contract, so this is not a regression — but baking the result into every page at build time makes the silence more durable. A hard failure when the file is absent, or a build-time assertion on the expected bundle count, would make a real regression visible.

sites/brc-analytics/services/workflows/hooks/UseEntities/utils.ts
```diff
@@ -17,16 +16,13 @@ import { LOGAN_SEARCH } from "@repo/shared/workflow/loganSearch";
  */
 export const ensureEntitiesLoaded = createEntitiesLoader(
   async (config: SiteConfig): Promise<void> => {
-    // Load in parallel so the optional pangenome fetch adds no serial latency
-    // to the core workflows/entities load that every data page depends on.
     await Promise.all([
```
NoopDog, line 19: **low** — `loadPangenomes` was the only writer of a site-specific key into the shared entity store, so with it gone the `EntityRoute | (string & {})` widening in `packages/shared/services/workflows/types.ts` is dead and now only weakens typing: `findEntity("workflowz", id)` type-checks and silently returns `undefined` instead of failing at the call site.

Narrowing `EntityStoreKey` back to `EntityRoute` restores the check — a small follow-up this PR has earned.

packages/shared/services/staticGeneration/workflows/utils.ts
frano-m: Fixed in 12abf673 — the hardcoded \`true\` is gone rather than restated.

New \`packages/shared/workflow/featureFlags.ts\` owns the category-to-flag map. \`buildOrganismWorkflows\` no longer knows about feature flags at all, so the build side simply emits the organism-compatible set — which *is* the flag-inclusive superset by definition, with nothing to keep in step. The render side filters through the shared \`filterFlagGatedWorkflowCategories\`.

The drift guard you asked for is the flag-state record: \`useWorkflowCategoryFeatureFlags\` returns an exhaustive \`Record<WorkflowCategoryFeatureFlag, boolean>\`, so adding a second gated category is a compile error there until it is resolved. \`4c2a72d5\` adds a test asserting the build emits the gated \`ASSEMBLY\` category, and I re-verified on built output: \`ASSEMBLY\` is in \`__NEXT_DATA__\` but absent from SSR markup across the 1,975 organism pages.

packages/shared/views/OrganismView/components/Main/components/WorkflowsSection/workflowsSection.tsx
frano-m: Fixed in 12abf673 — this filter is now a call to the shared \`filterFlagGatedWorkflowCategories\`, which is the only place the category-to-flag rule exists. Flag state is resolved in one exhaustive record (\`useWorkflowCategoryFeatureFlags\`), so a new gated category cannot be silently missed here. Details on the build half in the reply on \`staticGeneration/workflows/utils.ts\`.

sites/brc-analytics/services/staticGeneration/organism/utils.ts
frano-m: Fixed in 4c2a72d5 — two suites, 12 tests (628 to 640).

\`tests/services/staticGeneration.pangenomes.test.ts\` covers keying by \`speciesTaxonomyId\`, absent file, non-array payload, per-worker memoization, and eviction on rejection. \`tests/services/staticGeneration.workflowCategories.test.ts\` covers the same for \`loadWorkflowCategories\` plus the site-config path resolution and the flag-inclusive superset the organism pages embed — so \`loadWorkflowCategories\`'s export is now justified by an importer, as you suspected it should be.

One structural change was needed to get there: the pangenome reader moved to its own module (\`organism/pangenomes.ts\`). Importing \`organism/utils.ts\` pulls in \`@brc/config/config\`, which transitively reaches \`next-mdx-remote\` and fails to parse under jest. The reader is a pure file read with no config dependency, so splitting it made it directly testable.

sites/brc-analytics/services/staticGeneration/organism/utils.ts
frano-m: Fixed in d033a358 — every read failure now fails the build; the ENOENT-reads-as-empty path is gone.

I checked this was safe before hardening rather than assuming: \`catalog/output/pangenomes.json\` is committed, the catalog build writes named outputs and never clears the directory, and the Dockerfile copies the repo before \`build-brc-db\`, so there is no build path where the file is legitimately absent. The \`#1341\` "may not exist" caveat is stale.

Your framing also pointed at the coupling itself. Since nothing fetches the served copy any more (see the reply on \`UseEntities/utils.ts\`), d033a358 drops the \`OPTIONAL_JSONS\` block from \`sync-api-brc-analytics.sh\` entirely — so there is no second consumer left to stay in step with, and the comment no longer has to ask anyone to. Also refreshed the \`apis/pangenome.ts\` JSDoc, which still described entity-store fetching.

sites/brc-analytics/services/workflows/hooks/UseEntities/utils.ts
frano-m: Fixed in 0fecbacf — \`EntityStoreKey\` is deleted rather than narrowed; \`store.ts\` and \`query.ts\` use \`EntityRoute\` directly. \`findEntity("workflowz", id)\` is a type error again. Confirmed no remaining references and that every store write already passes an \`EntityRoute\` literal.

frano-m:
Thanks — all six addressed, plus a rebase onto latest \`main\` (clean, no conflicts). Four commits on top:

| Commit | Finding |
|---|---|
| \`12abf673\` | the two mediums — single owner for the category feature-flag gate |
| \`d033a358\` | pangenome soft-fail + the out-of-diff \`sync-api\` entry |
| \`0fecbacf\` | \`EntityStoreKey\` widening |
| \`4c2a72d5\` | the missing reader tests |

Detail is in the inline replies. Two things that belong here rather than on a line:

**\`scripts/sync-api-brc-analytics.sh:14\` (out of diff)** — dropped the entry. Confirmed nothing reads \`/api/pangenomes.json\` now that the loader, getter and route constant are gone; the served dir is gitignored and \`rm -f "$API_DIR"/*.json\` clears the stale copy on the next sync. Removing it also resolved the other half of your \`organism/utils.ts:8\` point: with no second consumer, there is no coupling left to document, and the build-time read now hard-fails.

**Judgment call #2** — taking your description as the record, since it is sharper than mine was: if the entity cache fails to load, the organism page renders perfectly while every "Configure Inputs" link dead-ends on the gated analyze page's error screen. It moves where the user learns something is wrong rather than removing the signal. No code change.

**One thing I deliberately did not do.** The same ASSEMBLY-skip rule also lives in \`buildAssemblyWorkflows\` and \`WorkflowsView\`'s \`getWorkflows\` — both pre-existing, both outside this diff. \`packages/shared/workflow/featureFlags.ts\` is the obvious home for them and the move is mechanical, but it is scope this PR did not create, so I left it for a follow-up rather than widening the diff. Say the word if you would rather it landed here.

**Verification** — 640 unit tests (was 628), \`tsc --noEmit\`, ESLint, Prettier all clean; both site builds compile. Re-checked flag semantics on real build output after the refactor: across the 1,975 BRC organism pages the gated \`ASSEMBLY\` category is in \`__NEXT_DATA__\` and absent from SSR markup, and the pangenome bundle is still embedded (spot-checked taxon 5855). Behaviour unchanged.

sites/brc-analytics/services/staticGeneration/organism/pangenomes.ts
```diff
@@ -0,0 +1,36 @@
...
+const PANGENOMES_STATIC_LOAD_FILE = "catalog/output/pangenomes.json";
+
+let pangenomesPromise: Promise<Map<string, Pangenome>> | null = null;
+
+/**
+ * Reads the pangenomes catalog into a map keyed by species taxonomy ID,
+ * memoized per build worker so the per-page lookup is O(1). Every failure —
+ * an absent file, a malformed payload, a payload that is not an array — fails
+ * the build, so a regression cannot silently ship every organism page without
+ * its pangenome section.
```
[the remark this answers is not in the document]
frano-m: Good catch on the inconsistency — the description was stale, not the code. Updated the PR body.

The hard failure is deliberate, from review feedback on the previous round: an absent catalog read as empty meant a renamed or missing output would ship all 1,975 organism pages with no Pangenome tab and no Pangenome section on a green build. That silence was tolerable when the data was fetched per visit and failed soft at runtime, but it becomes durable once the bundle is baked into every page at build time.

Checked it was safe before hardening rather than assuming: `catalog/output/pangenomes.json` is committed, the catalog build writes named outputs and never clears the directory, and the Dockerfile copies the repo before running `build-brc-db` — so there is no build path where the file is legitimately absent. The `#1341` "may not exist yet" caveat that justified the soft read is stale. Tests in `tests/services/staticGeneration.pangenomes.test.ts` assert the ENOENT case now rejects.

sites/brc-analytics/services/staticGeneration/organism/pangenomes.ts
```diff
@@ -0,0 +1,36 @@
+import { type Pangenome } from "@brc/apis/pangenome";
+import { promises as fsp } from "fs";
+
+// The pangenomes catalog, read at build time and embedded into each organism
+// page. Repo-root-relative: site builds run from the repo root.
+const PANGENOMES_STATIC_LOAD_FILE = "catalog/output/pangenomes.json";
```
NoopDog, line 8: **Every brc build is now hard-coupled to a hand-committed mock file.** No catalog build step produces `catalog/output/pangenomes.json` (`build-catalog.ts` writes only assemblies/organisms/outbreaks/workflows/mappings; the JSON was hand-committed in #1442), and this loader makes any read failure fatal to every organism page's `getStaticProps` — while this PR also deletes the sync script's optional handling that acknowledged the file "may not exist before its catalog build lands (#1341)".

Any pipeline rework or clean `catalog/output` regeneration that drops the mock makes every `next build` of the site fail with ENOENT — a hard release-pipeline outage for data the deleted runtime loader deliberately treated as optional. The hardcoded path is also a second, config-invisible copy of the repo-root cwd assumption: every other catalog source travels via a site-config `staticLoadFile` entry (the convention the sibling `getWorkflowsStaticLoadFile` relies on).

Suggestion: carry the pangenomes path via site config like the other catalogs, or — while the data is mock-backed — degrade to an empty map with a loud warning instead of failing the build.

packages/shared/workflow/featureFlags.ts
```diff
@@ -0,0 +1,66 @@
...
+export type WorkflowCategoryFeatureFlags = Record<
+  WorkflowCategoryFeatureFlag,
+  boolean
+>;
+
+/**
+ * The single owner of which feature flag gates which workflow category.
+ * Categories absent from this map are never gated.
+ */
+const FEATURE_FLAG_BY_CATEGORY: Partial<
```
NoopDog, line 34: **The "single owner" claim isn't true yet — the gating rule still exists in three copies.** The identical ASSEMBLY gate remains inline in `packages/shared/views/WorkflowsView/utils.ts:112` and `packages/shared/views/AnalyzeWorkflowsView/components/Main/utils.ts:30`, with raw `useFeatureFlag("assembly-workflows")` string literals in `workflowsView.tsx:22` and `AnalyzeWorkflowsView/components/Main/main.tsx:17`. `isWorkflowCategoryEnabled` is also a dead export (no consumers outside this module), and `filterFlagGatedWorkflowCategories` has no test coverage.

Consequence: adding the next flag-gated category to this registry (its advertised compile-error mechanism) gates only the organism page — the other two views keep their hard-coded ASSEMBLY-only checks and ship the new category ungated, exactly the drift this module was built to prevent.

Suggestion: migrate those two views to `filterFlagGatedWorkflowCategories`/`useWorkflowCategoryFeatureFlags` in this PR, or trim the registry's claims (and the dead export) to what it actually owns.

sites/brc-analytics/services/staticGeneration/organism/pangenomes.ts
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * Reads the pangenomes catalog into a map keyed by species taxonomy ID,
+ * memoized per build worker so the per-page lookup is O(1). Every failure —
+ * an absent file, a malformed payload, a payload that is not an array — fails
+ * the build, so a regression cannot silently ship every organism page without
+ * its pangenome section.
+ * @returns Pangenomes keyed by species taxonomy ID.
+ */
+export function loadPangenomes(): Promise<Map<string, Pangenome>> {
+  pangenomesPromise ??= fsp
```
NoopDog: **Duplicate memoized catalog reader.** This re-implements, nearly line for line, the memoized read/parse/`Array.isArray`-guard/evict-on-rejection reader this same PR adds as `loadWorkflowCategories` (`packages/shared/services/staticGeneration/workflows/utils.ts:35`) — a pattern `seedDatabase` already carries a third time — and each copy gets its own near-identical test suite (the "Don't cache a failed read" comment appears verbatim in both new files).

A fix to the eviction race or error shape in one copy won't reach the others. Suggestion: extract one shared `readCatalogArray<T>(path)` — the Map-keyed memo in `workflows/utils.ts` already supports arbitrary paths — so this file collapses to a map-by-`speciesTaxonomyId` post-step and the caching behavior is tested once.

sites/ga2/services/staticGeneration/organism/utils.ts
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * Attaches the build-computed fields the organism detail page renders — the
+ * organism's compatible workflow categories — so the page prerenders fully
+ * without the client entity store.
+ * @param organism - Organism record.
+ * @returns Organism detail data.
+ */
+export async function augmentOrganismDetail(
+  organism: GA2OrganismEntity
```
NoopDog: **Per-site duplication of the organism-detail augmenter.** ga2's `augmentOrganismDetail`/`GA2OrganismDetail` are a strict subset of brc-analytics' `augmentOrganismDetail`/`BRCOrganismDetail` — the shared `workflowCategories` half of the augmentation is duplicated per site, so the next shared build-computed field (or a serialization fix like the undefined-field handling brc's copy documents) must be hand-edited in both sites' `utils.ts` and `types.ts`, and missing one lets the two sites' prerendered pages silently drift.

Suggestion: move the generic piece to `packages/shared` next to `loadOrganismWorkflowCategories` — e.g. `WithWorkflowCategories<T>` plus `augmentWithWorkflowCategories(config, organism)` — and have brc compose only the pangenome on top; ga2's two files disappear.

packages/shared/services/staticGeneration/workflows/utils.ts
```diff
@@ -0,0 +1,68 @@
+import type { SiteConfig } from "@databiosphere/findable-ui/lib/config/entities";
+import type { OrganismContract } from "@repo/shared/apis/types";
+import type { WorkflowCategory } from "@repo/shared/apis/workflow";
+import { buildOrganismWorkflows } from "@repo/shared/views/OrganismView/components/Main/utils";
```
NoopDog: **Layering inversion: build-time services importing view internals.** This static-generation module imports domain logic (`buildOrganismWorkflows`) from `views/OrganismView/components/Main/utils`, so the build pipeline now depends on a view component's internals — relocating or refactoring the OrganismView tree breaks the build.

Suggestion: move `buildOrganismWorkflows` into the shared domain layer (`packages/shared/workflow/`, where this PR already placed `filterFlagGatedWorkflowCategories`) and import it from there in both the view and this service.

sites/brc-analytics/services/staticGeneration/organism/utils.ts
```diff
@@ -0,0 +1,29 @@
...
+ * Attaches the build-computed fields the organism detail page renders — the
+ * organism's compatible workflow categories and (when present) its species
+ * pangenome — so the page prerenders fully without the client entity store.
+ * @param organism - Organism record.
+ * @returns Organism detail data.
+ */
+export async function augmentOrganismDetail(
+  organism: BRCDataCatalogOrganism
+): Promise<BRCOrganismDetail> {
+  const workflowCategories = await loadOrganismWorkflowCategories(
```
NoopDog: **Independent loads awaited sequentially.** `loadOrganismWorkflowCategories` and `loadPangenomes` are independent, but awaited one after the other — the deleted client code ran the equivalent pair under `Promise.all` precisely "so the optional pangenome fetch adds no serial latency". This runs once per organism page (~2K pages) in `getStaticProps`; on each worker's first page the two file read+parses serialize instead of overlapping.

```ts
const [workflowCategories, pangenomes] = await Promise.all([
  loadOrganismWorkflowCategories(config, organism),
  loadPangenomes(),
]);
```

packages/shared/services/staticGeneration/workflows/utils.ts
```diff
@@ -0,0 +1,68 @@
...
+  return promise;
+}
+
+/**
+ * Returns the workflows catalog file path from the site's workflows entity
+ * config, so build and runtime read the same source.
+ * @param config - Site config accessor (provides the site's entity configs).
+ * @returns Workflows catalog file path.
+ */
+function getWorkflowsStaticLoadFile(
```
NoopDog, line 86: **Fourth copy of the route→entity-config lookup, re-run per page.** `getWorkflowsStaticLoadFile` re-runs `config().entities.find(({ route }) => route === "workflows")` on every per-page call, duplicating the lookup in `staticPaths.ts:21`, `staticProps.ts`, and `priority-pathogens/index.tsx`, keyed on a magic route string nothing ties back to the site configs — and its missing-file error shape already differs from `seedDatabase`'s.

Suggestion: extract a shared `findEntityConfig(config, route)` (or accept the `staticLoadFile` as an explicit parameter from the site's augmenter) and resolve it once per build instead of per page.

packages/shared/services/staticGeneration/workflows/utils.ts
```diff
@@ -0,0 +1,68 @@
+import type { SiteConfig } from "@databiosphere/findable-ui/lib/config/entities";
+import type { OrganismContract } from "@repo/shared/apis/types";
+import type { WorkflowCategory } from "@repo/shared/apis/workflow";
+import { buildOrganismWorkflows } from "@repo/shared/views/OrganismView/components/Main/utils";
+import { promises as fsp } from "fs";
+
+const categoriesByFile = new Map<string, Promise<WorkflowCategory[]>>();
```
NoopDog, line 8: **Nit (dev-only): process-lifetime memo serves stale catalogs in `next dev`.** `categoriesByFile` (and `pangenomesPromise` in the brc pangenomes reader) cache the catalog read with no invalidation, so after regenerating `catalog/output/*.json` a long-lived dev server keeps serving the old workflow categories/pangenome until restart — data the pre-PR runtime fetch picked up on every reload.

Mitigating: this matches the pre-existing `seedDatabase` convention ("immutable for the lifetime of a build"), which already made the organism record itself stale in dev. Worth at most a comment or a `NODE_ENV`-gated bypass.

packages/shared/services/staticGeneration/entity/staticProps.ts
```diff
@@ -18,7 +17,41 @@ export function makeEntityStaticProps<R>(
...
+  config: () => Pick<SiteConfig, "entities">,
+  entityListType: string,
+  meta: PageMeta,
+  augmentData: (data: R) => Promise<D>
+): GetStaticProps<EntityPageProps<D>, EntityPageParams>;
+/**
+ * Implementation. The site is decoupled by injecting its config accessor;
+ * when the entity's detail config opts into static loading, the entity record
+ * is fetched into `data` and augmented when an augmenter is supplied.
+ * @param config - Site config accessor (provides the site's entity configs).
```
NoopDog, line 14: **Overload triplication for two call sites.** `makeEntityStaticProps` grows two overload signatures plus the implementation signature — three near-identical parameter lists and three JSDoc blocks (~55 lines) — where the single signature `<R, D extends R = R>(…, augmentData?)` expresses the same contract for every current caller. Every future parameter change must now be edited in three places that can drift (the first overload's JSDoc already tells a narrower story than the implementation's).

The overloads' only extra guarantee — a caller can't claim an augmented `D` while omitting the augmenter — protects against a misuse no call site is near; if that guard isn't deliberately wanted, consider collapsing to the single generic signature.

NoopDog (commented):
Automated code review (high effort): 9 verified findings posted as inline comments — 1 correctness/robustness, 8 cleanup/consolidation. The one worth resolving before merge is the pangenome mock-file build coupling. (A 10th finding — the catalog regeneration that remaps seven organisms' taxids and 404s their old URLs — turned out to belong to the separately merged #1695, not this PR, so it is not posted here.)

sites/brc-analytics/services/staticGeneration/organism/pangenomes.ts
frano-m: Fixed in 5f3e6f3f — degraded rather than routed through config.

`loadPangenomes` now warns and returns an empty map on any unreadable or unusable catalog, so a clean `catalog/output` or a pipeline rework can no longer take the build down for an optional section.

On the config route: we decided against it. Config `entities` describe the site's *pages*, and pangenomes have no page — they're a section of the organism page. Adding an entity for them would widen the interface to carry something that isn't an entity. That reasoning is now recorded at the constant so the next reader doesn't re-litigate it:

```ts
// Not a site config entry: config entities describe the site's pages, and
// pangenomes have no page of their own -- they are a section of the organism page.
```

Two follow-on effects worth noting: the evict-on-rejection logic is gone (the promise always resolves now), and the degraded empty map is memoized for the build — an absent catalog stays absent, so retrying per page would buy ~2K failed reads and nothing else. The three tests that asserted build failure now assert the empty map *and* that the warning fires.

packages/shared/workflow/featureFlags.ts
frano-m: You're right, and the claim is trimmed in c4dd6e96 rather than made true here — taking the second half of your suggestion.

Confirmed both parts of your reading: `filterFlagGatedWorkflowCategories` has exactly one consumer (`workflowsSection.tsx`), and `isWorkflowCategoryEnabled` had no consumer outside its own module. The latter is now private.

The JSDoc no longer claims single ownership and names what escapes it:

> Not yet the only place the rule lives: `getWorkflows` and `buildAssemblyWorkflows` still carry their own inline checks, so a category added here is gated on the organism page alone until those views resolve their flags through this module too.

The exhaustiveness note is narrowed too — it holds wherever the record is *built*, which is the honest scope.

The migration is #1703, which also covers a fourth gate your comment didn't reach: `shouldIncludeWorkflow` gates Hyphy by matching a hardcoded TRS-ID prefix, so workflow-level gating has no registry at all. The plan there is a pure rule (build-time code can't call hooks — `buildOrganismWorkflows` now runs in `getStaticProps`) plus one `useWorkflowFeatureFlags` that owns resolution and binding, replacing the loose `= false` boolean params. Worth flagging the failure direction you identified: those defaults make the same rule fail *open* in two views and *closed* if an argument is forgotten. The flag-name literals are #1702.

sites/brc-analytics/services/staticGeneration/organism/pangenomes.ts
frano-m: Not addressed — flagging it rather than quietly leaving it.

The degrade fix (5f3e6f3f) has moved the two apart in the meantime: `loadPangenomes` no longer evicts on rejection or rethrows, because it now always resolves to a map. What's left in common is read → parse → `Array.isArray` guard → memoize. Still duplication, but the error-handling half that made a shared fix valuable is now genuinely different between the two, since one degrades and the other is build-fatal.

A shared `readCatalogArray<T>(path)` would have to take the failure policy as a parameter to serve both, which is most of what distinguishes them. Happy to do it if you still want it — say the word and I'll file it — but I'd rather not fold a third refactor into this PR to unify two readers that have just diverged on purpose.

sites/ga2/services/staticGeneration/organism/utils.ts
frano-m: Fixed in 5f3e6f3f — ga2's two files are gone.

`WithWorkflowCategories<T>` now lives in `packages/shared/services/staticGeneration/workflows/types.ts`, and the augmentation is shared. brc composes only the pangenome on top. ga2's page passes the shared augmenter directly.

One thing the sketch missed: `site-config/ga2/local/entity/organism/organismMainColumn.ts` also referenced `GA2OrganismDetail`, so that's repointed too — deleting the type without it would have broken the ga2 build.

The binding is a small curry, `makeWorkflowCategoriesAugmenter(config)`, rather than a bare `augmentWithWorkflowCategories(config, organism)`. That keeps the ga2 call site free of an inline arrow, and it gave the natural home for resolving `staticLoadFile` once — see the thread on the per-page lookup.

packages/shared/services/staticGeneration/workflows/utils.ts
frano-m: Fixed in 5f3e6f3f. `buildOrganismWorkflows` now lives at `packages/shared/workflow/organismWorkflows.ts`, beside `filterFlagGatedWorkflowCategories` as you suggested.

It turned out cleaner than a symbol move: `views/OrganismView/components/Main/utils.ts` contained only `buildOrganismWorkflows` and its private `workflowIsCompatibleWithOrganism` helper, and nothing in the view imported it — the service and its test were the only consumers. So the whole file moved via `git mv` (history preserved) and the directory is gone. The build pipeline no longer reaches into the view tree at all.

sites/brc-analytics/services/staticGeneration/organism/utils.ts
frano-m: Fixed in 5f3e6f3f, with the reason recorded at the call site so it doesn't regress:

```ts
// Read together: the two catalogs are independent, and this runs once per
// organism page, so serializing them would add the pangenome read's latency
// to every worker's first page for nothing.
const [withWorkflowCategories, pangenomes] = await Promise.all([
  augmentWorkflowCategories(organism),
  loadPangenomes(),
]);
```

packages/shared/services/staticGeneration/workflows/utils.ts
frano-m: Half fixed in 4de2aca2 — the per-page re-run is gone; the four-copy dedup isn't done.

`makeWorkflowCategoriesAugmenter` now resolves `staticLoadFile` once and keeps it in the closure, and `loadOrganismWorkflowCategories` takes the resolved path instead of the config. So the lookup runs once per augmenter rather than once per generated page.

Resolution is lazy rather than at bind time, deliberately: the augmenter is constructed at module scope, and resolving there would read the config at import time and turn a missing entity into an import-time throw. There's a test asserting the accessor is called exactly once across three pages.

On the shared `findEntityConfig(config, route)`: agreed in principle, but it reaches `staticPaths.ts`, `staticProps.ts` and `priority-pathogens/index.tsx`, none of which this PR otherwise touches, and the magic `"workflows"` route string is a broader config question. I'd rather file it than grow this PR — confirm and I'll open it.

packages/shared/services/staticGeneration/workflows/utils.ts
frano-m: Acknowledged, not changed — and the second option is one we can't take.

A `NODE_ENV`-gated bypass would be dev-vs-prod branching in app code, which we've ruled out as a matter of course: it makes dev and prod take different paths through the same module, which is how the class of bug that only reproduces in one of them gets in. So the choice was a comment or nothing.

Going with nothing, for now: as you note it matches the existing `seedDatabase` convention, which already makes the organism record itself stale in dev, so a comment on one of three readers would document a repo-wide behaviour in an arbitrary place. If it bites anyone in practice, the right fix is an invalidation story for all the build-time catalog readers at once, not a bypass in this one.

packages/shared/services/staticGeneration/entity/staticProps.ts
frano-m: Reverted in 5f3e6f3f — collapsed to the single `<R, D extends R = R>` signature, ~55 lines down to one JSDoc block.

Context you should have, because it'll come back: **those overloads exist because Copilot asked for them.** It flagged that the single signature lets a caller claim an augmented `D` while omitting `augmentData`, leaving the runtime value un-augmented behind a widened type; the overloads landed in response. Copilot has already re-raised it on this PR since the revert.

Going with your call — the misuse is one no call site is near, and three drifting parameter lists is the more likely cost. Noting the disagreement here so the next person to read the file finds the history rather than re-adding them.

tests/services/staticGeneration.pangenomes.test.ts
```diff
@@ -0,0 +1,93 @@
...
+    readFile.mockRejectedValue(error);
+
+    await expect(loadPangenomes()).resolves.toEqual(new Map());
+    // Loud: the section disappears from every organism page, so a silent
+    // degrade would ship unnoticed.
+    expect(warn).toHaveBeenCalledWith(
+      expect.stringContaining(PANGENOMES_STATIC_LOAD_FILE),
+      error
+    );
+  });
```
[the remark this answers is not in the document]
frano-m: Deliberate, not an oversight — this is the fix for the review finding above.

Making the catalog build-fatal is what the previous revision did, and it hard-couples every BRC build to a hand-committed mock: no catalog build step writes `catalog/output/pangenomes.json`, so any pipeline rework or clean `catalog/output` regeneration turns every `next build` into ENOENT. That's a release-pipeline outage for an optional page section.

The silent-shipping risk you're pointing at is real and handled by the warning rather than by the failure: the degrade logs loudly, and the tests assert the warning fires, not just the empty map. An organism page without its pangenome section is the page the site served before that section existed.

packages/shared/services/staticGeneration/entity/staticProps.ts
```diff
@@ -8,17 +8,21 @@ import type { EntityPageParams, EntityPageProps } from "./types";
...
  */
-export function makeEntityStaticProps<R>(
+export function makeEntityStaticProps<R, D extends R = R>(
   config: () => Pick<SiteConfig, "entities">,
   entityListType: string,
-  meta: PageMeta
-): GetStaticProps<EntityPageProps<R>, EntityPageParams> {
+  meta: PageMeta,
+  augmentData?: (data: R) => Promise<D>
+): GetStaticProps<EntityPageProps<D>, EntityPageParams> {
```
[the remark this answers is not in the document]
frano-m: Deliberate — the overloads were removed in response to a review finding on this PR, which called them ~55 lines of triplication across three parameter lists that can drift, guarding a misuse no call site is near.

The footgun you describe is accurate, and both signatures are defensible; this is a judgement call that's been made the other way. Leaving the reasoning in the thread on `staticProps.ts:43` so it isn't silently re-litigated.

sites/brc-analytics/services/staticGeneration/organism/pangenomes.ts
frano-m: Adding context that landed after the above: the third copy is scheduled for deletion.

findable-ui 56.0.0 ships a cached `seedDatabase` (DataBiosphere/findable-ui#987/#994), and our upgrade ticket #1698 carries "a decision recorded on our local `seedDatabase` vs the new upstream one" as an explicit acceptance criterion — the likely outcome being that `packages/shared/utils/seedDatabase/utils.ts` is dropped in favour of upstream's.

That changes the arithmetic on this thread. Unifying the three readers now would mean designing a shared abstraction around a module that is about to be deleted, and `seedDatabase` is the poorest fit of the three anyway: it seeds findable-ui's `database` singleton and does `Object.values()` rather than expecting an array, so it shares the memo-and-evict shape but not the read-and-parse.

That leaves the two array readers, which have themselves diverged since your comment — one degrades to an empty map with a warning, the other stays build-fatal — so a shared `readCatalogArray<T>` would have to take the failure policy as a parameter, which is most of what separates them.

So: deferring deliberately rather than declining. The right moment is after the v56 upgrade, when it's clear whether our `seedDatabase` survives at all. If it doesn't, the remaining duplication is two readers with different failure policies, and worth re-judging on that basis rather than on three.

frano-m:
## Suppressed Copilot comments — audit

Went through every suppressed comment across all four Copilot reviews on this PR (14 in total). They collapse into two distinct claims, both accounted for.

### 1. Pangenomes should be build-fatal — 13 comments, deliberately not actioned

Five on `pangenomes.ts` (impl + JSDoc) and eight on `staticGeneration.pangenomes.test.ts`, restating one objection: the reader degrades to an empty map instead of failing the build.

This is the fix for the review finding on `pangenomes.ts` above, not an oversight. Hard-failing is what the previous revision did, and it coupled every BRC build to a hand-committed mock: no catalog build step writes `catalog/output/pangenomes.json` (it was hand-committed in #1442), so any pipeline rework or clean `catalog/output` regeneration turns every `next build` into ENOENT — a release-pipeline outage for an optional page section.

The silent-shipping risk Copilot identifies is real and is handled by the warning rather than by the failure: the degrade logs loudly, and the tests assert the warning fires, not merely that the map is empty.

**One of these was a legitimate hit, though, and it's now fixed.** Several comments justified themselves by citing "the PR description's stated intent" — and they were right that the code and the description disagreed. The description was stale, still describing the pre-review revision ("Every read failure ... fails the build", and "Pangenome soft-fail (low). Now hard-fails"). The description has been corrected to describe the degrade and to note that hard-failing was tried and reverted in review. Copilot was reading a spec that no longer matched the decision.

### 2. Restore the `makeEntityStaticProps` overloads — 1 comment, deliberately not actioned

`staticProps.ts:25`, the same point Copilot raised inline. Answered on that thread: the overloads were removed in response to the review finding on `staticProps.ts:43`, which called them ~55 lines of triplication across three parameter lists that can drift, guarding a misuse no call site is near.

Worth recording plainly, since it will keep resurfacing: **Copilot and this review want opposite things here.** Both positions are defensible; the call has been made for the single signature, and the reasoning is in the `staticProps.ts:43` thread so the next person to touch the file finds the history rather than re-adding them.

### Net

No code changes came out of this audit — one documentation fix (the PR description). Everything else was already decided, with the reasoning recorded on the relevant thread.

packages/shared/services/staticGeneration/entity/staticProps.ts
```diff
@@ -8,17 +8,21 @@ import type { EntityPageParams, EntityPageProps } from "./types";
...
+ * an augmenter is supplied, given the build-computed fields the page renders
+ * (e.g. a pre-filtered workflows slice) so it can prerender fully.
  * @param config - Site config accessor (provides the site's entity configs).
  * @param entityListType - Entity list type (route segment) for the page.
  * @param meta - Page metadata (description and title).
+ * @param augmentData - Optional build-time augmenter applied to the record.
  * @returns getStaticProps.
  */
-export function makeEntityStaticProps<R>(
+export function makeEntityStaticProps<R, D extends R = R>(
```
NoopDog, line 20: **medium — the overloads that made `augmentData` mandatory for `D != R` were dropped again**

With `augmentData` optional but `D extends R = R` still explicitly claimable, the fallback on line 46 (`props.data = augmentData ? await augmentData(data) : (data as D)`) is an unsound cast: a caller can name a wider `D` and skip the augmenter, and `data as D` papers over it.

The overload pair added in 399cad7f closed exactly this hole, but 5f3e6f3f ("address review") removed it again — looks unintentional, since the PR body still lists it as fixed.

Concrete failure: the follow-up assembly-detail slice writes

```ts
makeEntityStaticProps<Assembly, AssemblyDetail>(config, "assemblies", META)
```

and forgets the fourth argument. It compiles. `props.data` ships with no `workflowCategories`, and `WorkflowsSection` calls `filterFlagGatedWorkflowCategories(undefined, flags)` → `TypeError: Cannot read properties of undefined (reading 'filter')` on every generated page.

Restoring the two overloads turns that into a compile error at the call site.

sites/brc-analytics/services/staticGeneration/organism/pangenomes.ts
```diff
@@ -0,0 +1,43 @@
...
+    .readFile(PANGENOMES_STATIC_LOAD_FILE, "utf8")
+    .then((text) => {
+      const pangenomes = JSON.parse(text) as Pangenome[];
+      if (!Array.isArray(pangenomes))
+        throw new Error(
+          `Pangenomes catalog is not an array: ${PANGENOMES_STATIC_LOAD_FILE}`
+        );
+      return new Map(pangenomes.map((p) => [p.speciesTaxonomyId, p]));
+    })
+    .catch((error) => {
```
NoopDog: **low — the `.catch` swallows corruption as well as absence**

The soft-fail rationale in the JSDoc is right about *absence*: no catalog build step writes this file yet, so a clean `catalog/output` legitimately has nothing to read, and failing the build over an optional section would be a bad trade. But the same `.catch` also absorbs a catalog that is present and unusable — including the `Array.isArray` guard three lines above, which throws and is then immediately caught.

Concrete failure: a truncated write leaves invalid JSON in `catalog/output/pangenomes.json`, or the #1341 catalog build lands emitting the schema-shaped `{"pangenomes": [...]}` object that `catalog/source/pangenomes.yml` and `catalog/schema/generated/pangenomes.json` describe rather than the bare array this reader expects. The build stays green, all ~1,975 organism pages ship without the Pangenome tab and section, and the only signal is one `console.warn` per build worker buried in the build log.

Suggest splitting the two cases: soft-fail on `ENOENT` (the case the comment actually argues for) and rethrow on parse/shape failures, e.g.

```ts
.catch((error) => {
  if ((error as NodeJS.ErrnoException).code !== "ENOENT") throw error;
  console.warn(...);
  return new Map<string, Pangenome>();
});
```

NoopDog (commented):
Code review pass over the branch. Verified locally at the branch head: `tsc --noEmit` clean, eslint clean, prettier clean, and the PR-relevant jest suites pass (the failures in a full run are pre-existing `ky`/ESM env artifacts unrelated to this diff).

I also checked the things that could plausibly have broken and found them sound, so no need to re-verify: `useFeatureFlag` uses `useSyncExternalStore` with a `false` server snapshot, so un-gating can't introduce a flag hydration mismatch; the build-time read of `catalog/output/workflows.json` is byte-equivalent to what `getWorkflows()` returned, and `loadWorkflows`' `extraWorkflows` only ever land in the by-id map, so nothing is lost by reading the catalog directly; every environment config for both sites carries a `workflows` entity, so `getWorkflowsStaticLoadFile` can't throw at build; the repo-root-relative catalog paths resolve because `next build sites/<site>` runs from the repo root; `buildOrganismWorkflows` has no other caller that depended on its removed flag argument; and nothing referenced the deleted `/api/pangenomes.json`.

Two findings inline — one medium, one low.

🤖 Reviewed with [Claude Code](https://claude.com/claude-code)

packages/shared/services/staticGeneration/entity/staticProps.ts
frano-m: Not addressed — this was deliberate, and it's your own earlier call on this PR.

The overloads went in at 399cad7f in response to Copilot. You then asked for them to be collapsed in https://github.com/galaxyproject/brc-analytics/pull/1680#discussion_r3891650436 — "~55 lines of triplication across three parameter lists that can drift, guarding a misuse no call site is near" — and 5f3e6f3f did exactly that. The reasoning and the disagreement are recorded in two threads so it wouldn't be re-litigated:

- https://github.com/galaxyproject/brc-analytics/pull/1680#discussion_r3900949869 — the reply to your comment, noting the overloads exist because Copilot asked for them and that Copilot had already re-raised it since the revert.
- https://github.com/galaxyproject/brc-analytics/pull/1680#discussion_r3900950068 — the reply to Copilot's re-raise, saying the footgun is real and the judgement has been made the other way.

So "looks unintentional, since the PR body still lists it as fixed" isn't right: the PR body records the overloads landing, and the collapse is a later, requested change. If the body reads as stale I'll fix the body.

On the substance: there's no live unsoundness. All three call sites — `priority-pathogens` (`<Outbreak>`, default `D = R`) and both organism detail pages (explicit `D`, augmenter supplied) — are correct today. The failure you describe is a call site in the follow-up assembly slice that doesn't exist yet, and it would be caught the first time that page is opened in a build.

Both signatures are defensible and I don't feel strongly. But this is now the third pass over the same trade-off with the answer flipping each time, so I'd rather it be settled explicitly than swapped again: if you want the compile-time guard back, say so as a reversal of r3891650436 and I'll restore the overloads and note on the thread that the call went the other way. Otherwise it stays as-is.

sites/brc-analytics/services/staticGeneration/organism/pangenomes.ts
frano-m: Fixed in c1fa7915 — split as suggested.

`ENOENT` warns and yields an empty map; everything else rethrows. That keeps the case your earlier comment (https://github.com/galaxyproject/brc-analytics/pull/1680#discussion_r3891650402) actually argued for — a clean `catalog/output` or a pipeline rework dropping the hand-committed file can't take the build down for an optional section — while a catalog that is present and wrong now fails loudly, which is the hole the degrade opened.

You're right that the shape case was the sharp one: `catalog/source/pangenomes.yml` is `{pangenomes: [...]}` and `catalog/schema/generated/schema.ts` types it that way, so the wrapped object is the *natural* thing for the #1341 build step to emit. The `Array.isArray` guard was catching it and the `.catch` three lines down was eating it — the guard was effectively dead code.

Also restored the evict-on-rejection the degrade had made moot, matching `loadWorkflowCategories`, so a transient read failure doesn't poison the rest of the worker. The absent-catalog path resolves rather than rejects, so it stays memoized and doesn't retry ~2K times.

Tests updated to match: absent still asserts empty map + warning, plus memoization of that path; non-array, invalid JSON and a non-`ENOENT` read error now assert a throw; and one case covers the eviction.

frano-m:
Re the suppressed comment on [this review](https://github.com/galaxyproject/brc-analytics/pull/1680#pullrequestreview-5085412165) — correct catch, and it was the description that was stale, not the code.

The reader's semantics changed in c1fa7915 in response to [r3910105250](https://github.com/galaxyproject/brc-analytics/pull/1680#discussion_r3910105250): degrade on `ENOENT` only, rethrow on malformed JSON / non-array payload. The "every read failure" wording dated from round 2, when the degrade was unconditional.

PR description updated to match: the How bullet now splits absent (warn + empty map) from present-but-unusable (build fails), and a new "Review round 3" section records both this and the two findings from the preceding review. Implementation and tests unchanged — the ENOENT-only split is the intended contract.

NoopDog (approved):

